Issue #16, opened 2014-03-12
All examples need descriptions and public domain notice


PaulStoffregen:
In a style similar to Arduino's examples, in File > Examples > 01.Basics

kenokenobingo:
What about adding a `LICENSE` file additionally to the repo? :)

PR #100, opened 2014-11-29
Asembler to copy data. 

FrankBoesing:


PaulStoffregen:
Any chance you'd consider releasing this as MIT license?

The rest of the audio lib is MIT.  If I include this, it forces GPLv3.

FrankBoesing:
No problem, but let me test this first (again)
It's quit old, an i'm not sure anymore that the code is ok.

FrankBoesing:
Ok, there's indeed a little bug, the sequence should be 	
pkhbt r3,r5,r6,LSL #16
pkhtb r4,r6,r5,ASR #16
pkhbt r5,r7,r8,LSL #16
pkhtb r6,r8,r7,ASR #16
...
I wrote a testsketch to verify the corrrectness.
(btw, the testbench made it possible to do a benchmark. The assembler-variant is ~3 times faster than the gcc-optimized copy: 274 us / 809 us for 100 calls of "memcpy_tointerlaceLR()" )

I clean up the code now, check the "mono"-variants too and replace the licence.
I close this pullrequest and do a new one this weekend.

PR #104, opened 2015-01-18
AnalysisRMS, fix in AnalysisPeak and aditional DAC control command

sgorpi:
I have added 2 functions (on/off) to control the DAC's volume ramping when using dacVolume. 
Furthermore, in reviewing the code of AnalyzePeak, I found it was using the difference between the positive and negative peak in a block, while normally, it'd be the maximum absolute distance from DC. Therefore, the current read() function has been renamed to readDiffPosNeg(), while the new read() function returns the maximum absolute distance from DC.
Lastly, I've added a class to compute the RMS of a block, based on the arm_math.h RMS. An example modified from PeakMeterStereo shows peak and RMS values of the audio.


sgorpi:
Now also added a Mid/Side encoder/decoder effect and the GUI is updated to include the additional components.

PaulStoffregen:
I've manually merged all this stuff.  Sorry it took so very long.

Thanks!

Issue #84, opened 2014-09-30
AudioEffectMultiply example needed


PaulStoffregen:
Is it possible to use this to create a 300 or 1200 baud modem signal?  That's be pretty awesome...

Some sort of useful example is really needed.

PR #105, opened 2015-01-18
Change max Q of SVF to 25

svgeesus:
Max Q was 5, too small for resonant filter effects common in analog synthesis. Increased to 25. Also updated GUI documentation to reflect this change.

Added example (which uses Q of 18) with LFO sweeping resonant frequency.
Example actually has two LFOs which are multiplied to give a more complex modulation.

This example also helps with issues https://github.com/PaulStoffregen/Audio/issues/86 and https://github.com/PaulStoffregen/Audio/issues/84


PaulStoffregen:
Did you verify the internal calculations don't overflow?

Nantonos:
the resonance function uss floats to calculate a damping factor
setting_damp = (1.0 / q) * 1073741824.0;

1 ÷ 0.707 × 1,073,741,824 = 1,518,729,595.4738331
1 ÷ 25 × 1,073,741,824 = 42,949,672.96

setting_damp is int32_t and the largest value it has to hold is with the lowest Q (clamped to 0.7).
setting_damp is used in filter_variable.cpp to set damp, also uit32_t which gets used like this

highpass = input - lowpass - MULT(damp, bandpass);
so it seems that multiplying by a smaller value of damp should not cause overflow.

I can see that changing the minimum Q value risks overflow somewhare, but not changing max Q.

PaulStoffregen:
Higher Q results in more internal gain at resonance.

Nantonos:
Yes, it would. But I am not seeing anywhere that damp is used to divide (where a smaller value would give higher numbers and risk overflow). I confess that surprises me, but looking at the code in AudioFilterStateVariable::update_fixed and AudioFilterStateVariable::update_variable I am not seeing anywhere that would overflow.

PaulStoffregen:
It's a tricky feedback loop.  I chose the max based on careful prototyping and testing the numerical ranges in all steps of the calculation.  Simply increasing the maximum without careful examination and testing of the numerical ranges makes me really nervous.

Nantonos:
I would be happy to look over your testing notes about numeric ranges. its entirely possible that I haven't understood where the code could overflow.

PaulStoffregen:
I don't have any "notes", but I do have this test code.

    #include <stdio.h>
    #include <stdlib.h>
    #include <stdint.h>
    #include <math.h>

    double fc = 820;   // max is 1/6th of sample rate
    double q = 5;      // allow range 0.7 to 5

    double hz = 870;
    double amplitude = 0.25;
    double samplerate = 48e6 / 1088.0;
    int length=190;
    double pi = 3.141592654;

    // http://www.musicdsp.org/showArchiveComment.php?ArchiveID=92

    static int32_t imult(int32_t a, int32_t b);

    int rs = 30;

    int main()
    {
        int i;
        double fmult, damp;
        double sample, prevsample=0;
        double low, band, high;
        double lowout, bandout, highout;
        int64_t ifmult, idamp;
        int32_t isample, iprevsample=0;
        int32_t ilow, iband, ihigh;
        int32_t ilowout, ibandout, ihighout;
        double  elow, eband, ehigh;
        double scale = 32768.0 * 4096;

        fmult = 2 * sin(pi * fc / (samplerate * 2));  // fmult is 0 to 1.0, for 0 to samplerate/6
        damp = 1 / q;

        ifmult = fmult * 65536.0 * 16384.0;
        idamp = damp * 65536.0 * 16384.0;

        // filter is stable when:  fmult*fmult + 2*damp < 4.0

        low = 0.0;
        band = 0.0;

        ilow = 0;
        iband = 0;

        for (i=0; i<length; i++) {
                sample = sin(pi*2.0 * hz * i / samplerate) * amplitude;

                printf("%6.3f", sample);

                // compute the filter at floating point
                low = low + fmult * band;
                high = (sample+prevsample)/2.0 - low - damp*band;
                prevsample = sample;
                band = band + fmult * high;
                lowout = low;
                highout = high;
                bandout = band;
                low = low + fmult * band;
                high = sample - low - damp*band;
                band = band + fmult * high;
    #if 1
                lowout = (lowout + low) / 2.0;
                bandout = (bandout + band) / 2.0;
                highout = (highout + high) / 2.0;
    #else
                lowout = lowout;
                bandout = bandout;
                highout = highout;
    #endif

                printf("  %6.3f, %6.3f, %6.3f", lowout, bandout, highout);

                isample = (int32_t)(sample * scale);
                printf("%9d", isample >> 12);

                // compute the filter at integers
                ilow = ilow + imult(ifmult, iband);
                ihigh = ((isample+iprevsample)>>1) - ilow - imult(idamp, iband);
                iprevsample = isample;
                iband = iband + imult(ifmult, ihigh);
                ilowout = ilow;
                ihighout = ihigh;
                ibandout = iband;
                ilow = ilow + imult(ifmult, iband);
                ihigh = isample - ilow - imult(idamp, iband);
                iband = iband + imult(ifmult, ihigh);
    #if 1
                ilowout = (ilowout + ilow) >> 1;
                ibandout = (ibandout + iband) >> 1;
                ihighout = (ihighout + ihigh) >> 1;
    #else
                ilowout = ilow;
                ihighout = ihigh;
                ibandout = iband;
    #endif

                // how much error is there between the intgers and floats?
                elow  = 100.0 * (lowout - (double)ilowout / scale) / lowout;
                eband = 100.0 * (bandout - (double)ibandout / scale) / bandout;
                ehigh = 100.0 * (highout - (double)ihighout / scale) / highout;

                printf("  %7d, %7d, %7d  (%.2f%%,%.2f%%,%.2f%%)",
                        ilowout >> 12, ibandout >> 12, ihighout >> 12, elow, eband, ehigh);




                printf("\n");



        }
        return 0;
    }
    static int32_t imult(int32_t a, int32_t b)
    {
        int64_t result;
        int32_t out;

        result = (int64_t)a * (int64_t)b;
    #if 1
        if (result & 0x80000000) result += (int64_t)0x0000000080000000;
        out = result >> 32;
        return out << 2;
    #else
        return result >> rs;
    #endif
    }

Nantonos:
Thanks for that. Computing the difference between an integer and a double implementation of the same filter is a nice way to explore for errors.

I see a comment in the code
// filter is stable when:  fmult*fmult + 2*damp < 4.0
As fmult ranges from 0 to 1, fmult*fmult will also range from 0 to 1. Taking 1 as the worst case, the filter is therefore stable when 2*damp < 3.0 or when damp < 1.5 or when q = 1/damp < 0.667
I guess this is where the 0.7 lower limit on Q comes from.

Running this code (gcc on cygwin) with the precision on the three error columns bumped up to three decimal places rather than 2, I see that Q can range down as low as 0.51 with that code; at 0.50 or below the errors go from all zeroes to thousands of percent.

At the higher end, with q=18 the errors are all zeroes as they are with Q=5. With Q = 25 there was one error of -0.012% and with Q = 24, a couple of errors 0.014% and -0.006%. These seem very small and unlikely to be audible.

At any rate, having run this test code, I remain convinced that the algorithm is not troubled by high Q values. It is troubled by low ones (depending on fmult, but in the range 0.5 to 0.7 is where it breaks).

This stability analysis is independent of clipping. As the documentation currently states, if Q is greater than 0.7 there will be gain at resonance and thus the input level may need to be reduced to avoid clipping. That is true at the current max of 5.0 and indeed at a Q of 1.0 even.

Another way to look at it is that as Q increases, the values further away from the resonant frequency are *lowered*.

PaulStoffregen:
I just ran the test, with Q changed from 5 to 25.  It clearly shows trouble.

Here's the last 6 lines:

    -0.181   1.977, -0.445, -2.140    -5918    64795,  -14569,  -70108  (0.00%,-0.00%,0.00%)
    -0.201   1.918, -0.690, -2.091    -6573    62857,  -22624,  -68522  (0.00%,-0.00%,0.00%)
    -0.217   1.831, -0.928, -2.011    -7127    59986,  -30404,  -65913  (0.00%,-0.00%,0.00%)
    -0.231   1.716, -1.153, -1.902    -7572    56217,  -37792,  -62312  (0.00%,0.00%,0.00%)
    -0.241   1.575, -1.363, -1.763    -7900    51599,  -44676,  -57765  (0.00%,0.00%,0.00%)
    -0.247   1.410, -1.555, -1.597    -8108    46194,  -50952,  -52335  (0.00%,0.00%,0.00%)

Notice how all 3 integer output have numbers beyond the int16_t range!

PaulStoffregen:
To see the problem more clearly, try changing amplitude to 0.95 (line 10).

    -0.504   7.630,  0.219, -8.137   -16522   250034,    7161,  257669  (0.00%,-0.00%,196.64%)
    -0.600   7.628, -0.737, -8.194   -19656   249957,  -24164,  255798  (0.00%,0.00%,195.27%)
    -0.686   7.514, -1.689, -8.130   -22488   246222,  -55362,  257877  (0.00%,0.00%,196.80%)

Now the highpass accumulator has numerical overflow, resulting in massive error.

To properly accommodate higher Q, the scaling of inputs and coefficients and outputs needs to be adjusted.

Nantonos:
Ah, I was only looking at the last three columns (the errors). I had not noticed that the integer part was using a wider int than the teensy code uses.

Will look at some more.

Nantonos:
Here are some lines also showing int16_t overflow
```
 0.034  -1.998, -0.145,  2.069     1118   -65459,   -4746,   67789  (0.000%,-0.000%,0.000%)
 0.151  -2.007,  0.103,  2.135     4954   -65778,    3366,   69952  (0.000%,0.000%,0.000%)
 0.266  -1.988,  0.355,  2.168     8714   -65144,   11638,   71045  (0.000%,0.000%,0.000%)
 0.377  -1.939,  0.609,  2.168    12341   -63542,   19940,   71028  (-0.000%,0.000%,-0.000%)
 0.482  -1.861,  0.859,  2.132    15778   -60973,   28141,   69876  (-0.000%,0.000%,-0.000%)
 0.579  -1.753,  1.102,  2.063    18974   -57453,   36108,   67586  (-0.000%,0.000%,-0.000%)
```
again with ampliture 0.95. But here, Q is 4.

PaulStoffregen:
Please understand my reluctance to merge this sort of change without extensive testing and verification.

These IIR filters seem deceptively simple, but they're really quite complex beasts.

My test code was never meant to be a comprehensive test or validation.  It was really just my own fiddling with the algorithm.  I probably recompiled and ran that test over a hundred times while writing the code we have now.

I do understand that higher Q is needed for synthesis applications.  This can and will happen, but it needs to be done with great caution.

Nantonos:
I have no problem with caution.

PaulStoffregen:
Yes, I mis-spoke in that earlier reply.  It's been a while since I've look at this algorithm.....

There's 2 types of overflow here.

PaulStoffregen:
If the filter is configured for gain and the input signal is too large, the filter is supposed to clip the output.  That happens by the signed_saturate_rshift() on lines 155 to 157.  https://github.com/PaulStoffregen/Audio/blob/master/filter_variable.cpp#L155

PaulStoffregen:
The problem we can't allow is overflow which occurs before signed_saturate_rshift(), which could be in any of the filter's other calculations, or in the addition right before that function is used.  If any of those overflow int32_t, the filter fails terribly.

PaulStoffregen:
To make Q=25 work, the left shift by 12 probably needs to be reduced, maybe to 10 or 9 or even 8.  The final signed_saturate_rshift() would be reduced by the same number of bits.  This will buy additional range, at the expensive of small round-off errors accumulating.

Computation of the filter coefficients also would need to be adjusted.

PaulStoffregen:
I was thinking about filter resonance and gain again this morning.

Something I originally considered, but didn't do, was using a scale factor on the output, rather the saturating rshift.  For modest gain, the input can be attenuated.  But doing too much of that leads to poor performance, because the original 16 bit material is truncated (without proper dithering) to much lower res, then filtered.

Scaling things well could allow the full resolution to be kept.  When higher Q is supported, this will probably become even more important.

The main issue before was the lack of a well defined equation for the peak gain at resonance.  I recall reading a few research papers on the subject, but none had a clear equation or algorithm.  But with some scripting, I believe the test program could be used to generate a large data set, which could then be 2 dimensional fitted to the resonance and corner frequency.

Nantonos:
That sounds like a much better approach. Requiring authors to take a guess at the input level is unsatisfactory (and is already required, even with the 0.7 to 5 Q range). It will tend to produce under estimates and thus lower quality. (If people guess too high, the hard clip is noticeable so they will lower it; if they guess too low, the resulting poorer quality is less noticeable).

Analysis of a large dataset to produce an autoscaling equation would be much more satisfactory from an ease of use point of view in addition to the higher quality.

I may be able to discuss this with some folks at the Web Audio conference next week. (Web Audio uses float32 rather than unit16 samples, and uses a single biquad filter node for all filtering, but the folk there are likely to have informed opinions on the issue).

svgeesus:
@PaulStoffregen I guess this one can be closed now

PR #113, opened 2015-05-01
BUG: Small error in linear interpolation

endolith:
This is from https://forum.pjrc.com/threads/27608-Very-minor-mistake-in-Audio-synth_sine-cpp

I haven't verified that it's correct myself.

Also see https://github.com/endolith/Audio/commit/210516665c394a836e76c79a4cce88daa2baecf6 where I combined these into a function (to make it easier to create new additive synthesis functions).  I can submit that too if it's a desirable change.

### DIFF
--- a/synth_sine.cpp
+++ b/synth_sine.cpp
@@ -50,7 +50,7 @@ void AudioSynthWaveformSine::update(void)
 				val2 = AudioWaveformSine[index+1];
 				scale = (ph >> 8) & 0xFFFF;
 				val2 *= scale;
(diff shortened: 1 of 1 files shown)


endolith:
(Somewhat related, I tried other fast sinusoid synthesis methods but didn't have much luck.  Resonant version is faster, but doesn't support all frequencies.  Quadratic version isn't any faster and is much more distorted: https://gist.github.com/endolith/669642b30eedb56c511e)

PR #115, opened 2015-05-31
Canceling is desirable for some, some of the time.

robsoles:
Harmless change to lower bound restriction of gain control per input of AudioMixer4 allows end users to attempt noise cancellation in terms of being able to use a second mic to cut down (or out, if possible) background/environment noise.

That's how I am calling it at least, inspiration is here: https://forum.pjrc.com/threads/28699-Little-audio-project?p=73830&viewfull=1#post73830

### DIFF
--- a/examples/Canceling/Canceling.ino
+++ b/examples/Canceling/Canceling.ino
@@ -17,7 +17,7 @@ AudioConnection Patch4(sine3000,0,mix1,3);
 
 AudioConnection Patch5(mix1,dac1);
 
-Void Setup()
+void setup()
 {
 	AudioMemory(6);
 	Serial.begin(9600);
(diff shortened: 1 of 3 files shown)


Pensive:
Once this is tested and proven working, let me know and I'll fix up the additional audiomixers i added. (8, 16 etc.)

Love the simplicity of this noise cancelling solution =D, it is elegant if not perfect.

robsoles:
The results of my testing seemed good enough or I would not have basically published, tho it is only right to wait for others to confirm or confirm for self.

I had knocked apart  copies of mixer.h/cpp and restructured it to do a simplistic shot at it, in the thread I indicate as inspiration, but I went back and looked at original mixer source files again and I was pretty stoked when I realised that changing two numeric values in one line of that code made something heaps better than the one I knocked up based on it - re-tested on this method and result appeared to be (easily) as good with less objects required.

I'm going to have to set up two microphones and play sometime - I've only tried this using virtual sources, a new Italian friend of mine is going to try it with actual microphones, and a real live scenario to need something like this, much sooner than I am going to get an opportunity to.

PR #119, opened 2015-08-09
Bitcrusher, Bitcrusher example, 2 channel mixer, 8 channel mixer, 16 channel mixer plus gui additions.

Pensive:
Sorry, my first github, thought my forked changes would autmoatically become a pull. 

so this is quite a bit of stuff in one pull.

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -65,6 +65,7 @@
 #include "analyze_peak.h"
 #include "control_sgtl5000.h"
 #include "control_wm8731.h"
+#include "effect_bitcrusher.h"
 #include "effect_chorus.h"
 #include "effect_fade.h"
 #include "effect_flange.h"
--- a/effect_bitcrusher.cpp
+++ b/effect_bitcrusher.cpp
@@ -1,0 +1,130 @@
+/* Audio Library for Teensy 3.X
+ * Copyright (c) 2014, Jonathan Payne (jon@jonnypayne.com)
+ * Based on Effect_Fade by Paul Stoffregen
+ * Also samplerate reduction based on Pete Brown's bitcrusher here: 
+ * http://10rem.net/blog/2013/01/13/a-simple-bitcrusher-and-sample-rate-reducer-in-cplusplus-for-a-windows-store-app
+ *
+ * Development of this audio library was funded by PJRC.COM, LLC by sales of
+ * Teensy and Audio Adaptor boards.  Please support PJRC's efforts to develop
+ * open source software by purchasing Teensy or other PJRC products.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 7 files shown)


PaulStoffregen:
Can you please submit this pull request without the 8 and 16 channel mixer objects?

Pensive:
Ok, but is there a way to do that without removing them from my fork? Can a pull request be stipulated on a line by line basis? I'm using them in 3 other projects and my workflow is already fairly convoluted.

May I ask why more capable mixer sets are not to be included? I'm sure you have good reasons, I'd like to understand.

Rgds

Jon

PaulStoffregen:
The reason involve memory usage.  A 16 channel mixer forces 16 inputs to reserve a buffer simultaneously.

Connecting 4 channel mixers in tandem causes each to force only 4 buffers allocated, which are mixed down to a single buffer and the other 3 returned to the memory pool.

The other issue involves delays associated with back connections.  When you create the audio connection objects, they have zero delay if they connect from an earlier to a later object, as defined by the order the object instances appear in your code.  Connections from a later to earlier object have a 2.9 ms delay.  Wide mixers might encourage a connection style to feed lots of to a huge mixer.  Of course, you can build that with many smaller mixers connected in tandem to make a big one, but when designing a system, smaller mixers encourage a style where of "localized" mixing where it's needed in the signal path... which leads to both better memory usage and avoiding latency/delay.

I'd be willing to add a 6 channel mixer, and maybe even 8, but 16 is far too much.

PaulStoffregen:
There's probably some complex set of git commands.... but at this point, probably the simplest thing is if I manually merge the code.  Will do that in a moment, then close this pull request.

PaulStoffregen:
I'm looking through the bitcrusher code now, and reformatting the white space to the style used by the rest of the library.

On the public API, the rest of the library uses functions named for the setting they access.  To be consistent with the audio lib (and Arduino's library style guidelines in general), the object should have 2 functions with names like sampleRate() and bits().

PaulStoffregen:
Ok, I've merged the Bitcrusher effect.  I made many tiny edits to fit the library's style.

Pensive:
OK thank you Paul

 

I'll review all your edits and changes to the bitcrusher for my own education;

 

If you're happy to, I'd like to do a new pull request for 2, 6 and an 8 channel mixer ( 8 channel mixer to be marked with a "High memory usage" warning in the gui, explaining the recommendation to use/stack smaller mixers for memory reasons).

I'll refactor my code to use 8 channel mixers instead, if I'm careful with my design this won't introduce any additional generation loss.

 

The reason I'm keen on bigger mixers is partly cleaner code, but mostly sound quality > Going through multiple mixers is multiple gain processes - this has a detrimental effect on sound quality. Typically with clean/effects mixers and mastering at the end to stereo, I end up with 5 mixer stages at 0.8 gain, by the time it comes out the end, an originally normalised waveform is emerging at 33% the volume it was at the start. I then need to boost it up, effectively losing 67% of the bitdepth detail.

 

But that's only an issue on really complex projects.

PaulStoffregen:
Sure, 2, 6 and 8 channel mixers are fine.

PR #126, opened 2015-12-07
Some bugfixes and some improvements

mamuesp:
- fixed: clear selected nodes, works with several or all selected nodes (reason: two counting variables with the same name "i" where used in two nested loops)
- new: change node names (you can assign names to the nodes, these will be used as IDs if exported or re-imported to CPP (Arduino code)
- fixed: import merges imported nodes to existings ones, skipping nodes with the same name
- fixed: tooltips are now working, extracting the first paragraph of the HTML code
- new: nodes definition extracted to JSON file
- new: nodes help pages extracted to single HTML files
- new: form pages extracted to single HTML files
- new: asynchronous loading of JSON, help and form data, data will only be loaded when needed
- style: adapted the size of the palette, labels had been partially hidden


PaulStoffregen:
The documentation panel is no longer working in Firefox when loading index.html (file:///tmp/Audio/gui/index.html) without a server.

PaulStoffregen:
This is probably due to the async loading.... we've tried that before and it doesn't work on most browsers when used offline.  This change was made specifically to enable completely offline usage on all browsers.  https://github.com/PaulStoffregen/Audio/commit/0e9d655c698e12b368802a571c7e40520025ece7

mamuesp:
I'll look into it - I have a solution in mind. 

Regards, Manfred

Sent from my iPad
Von meinem iPad gesendet

PaulStoffregen:
I'd really like to merge these ASAP.

* fixed: clear selected nodes, works with several or all selected nodes (reason: two counting variables with the same name "i" where used in two nested loops)
* new: change node names (you can assign names to the nodes, these will be used as IDs if exported or re-imported to CPP (Arduino code)
* fixed: import merges imported nodes to existings ones, skipping nodes with the same name
* fixed: tooltips are now working, extracting the first paragraph of the HTML code
* style: adapted the size of the palette, labels had been partially hidden

Any chance you could send a pull request with only those changes?

Or I could try to manually merge in just those changes, if you don't have the time to prepare another pull request.

PaulStoffregen:
I have mixed feeling about the async loading.  Of course the main problem is breaking offline usage.

Even if it can work offline, is this really best?  Loading one giant index.html doesn't take very long.  The file is only 122K.  Even if we eventually triple the number of library features, that's still a pretty reasonable size for today's internet, especially with servers & browsers that automatically deflate/gzip.

What I'm trying to say is perhaps the extra wait for initial loading is better than many smaller waits when each object's documentation is needed.  Today we get that for the images in the docs, but the html part appears instantly.  I'm not excited about adding interactive lag, even if it makes editing the documentation easier.

mamuesp:
I'll just fix the offline problem (not a simple rollback but an adaption so I may re-use the code later, perhaps we find a better solution ...) and so you may just merge the two (one will come ...) pull request in line, and all is done. Or you cancel this one an I create a complete new one when ready, this could be the easier way ...

Regards, Manfred

PaulStoffregen:
Looks like this broke both the export and import buttons.  :(

mamuesp:
Sorry, I was to fast - I didn't test the form loading well enough ... just a moment.

mamuesp:
Sorry again. I see I just forgot to adapt the forms as well ... I thought this was done ... a matter of a couple of minutes ...

PaulStoffregen:
Thanks.  :)

mamuesp:
done ... ooops ... ;-)

Issue #129, opened 2015-12-08
Import function


mamuesp:
For the moment, the import function is not very error tolerant. So it is necessary to enter the code between the to comment lines:
_// GUItool: begin automatically generated code_
_// GUItool: end automatically generated code_
and the lines must have a least two "words" separated by a whitespace - if there is no whitespace, the pattern matching regex will crash.
This will be fixed in a future version.

PaulStoffregen:
Any chance you might take a look at issue #59.  Is this sort of things practical to do?

Especially as the library grows with more types of hardware objects, conflicts for the fixed hardware resources are going to be come a really confusing problem.

mamuesp:
I believe this is not a big problem - the extension with the link to the hardware parameters and/or pins to read or write was one goal when I started looking into the code. So after I made my importer more tolerant (you should be able to throw "Moby Dick" in it without any error coming up ...) ;-), I will look into the mentioned issue.

My little goal is to be able to assemble a kind of "minimoog" with the tool - if that is possible, I believe also others would like to play along ...

Issue #59, opened 2014-09-10
GUI should mutually exclude certain objects (eg, I2S master vs slave)


PaulStoffregen:

PaulStoffregen:
I'd like to add an optional field to the JSON nodes definitions, so each object which uses hardware can specify a list of the hardware resources it requires, and what configuration is needs on each of piece of hardware.

For example, the "adc" object might look like this:

    "HardwareRequire":[
        {"hardware":"ADC0", "usage":"exclusive"},
        {"hardware":"PDB",  "usage":"44.1kHz"},
    ]

The "dac" object would look like this:

    "HardwareRequire":[
        {"hardware":"DAC0", "usage":"exclusive"},
        {"hardware":"PDB",  "usage":"44.1kHz"},
    ]

When a new object is dragged onto the canvas, its HardwareRequire list would be checked against all the objects currently on the canvas.  The word "exclusive" means the hardware resource can't be shared.  Any other word means that hardware can be shared with other objects, as long as they have the same usage.  If there's a conflict, the object would not be added to the canvas, and a (hopefully well worded) message would appear to explain the conflict.

For example, a feature I'm planning to someday try is 4X oversampling adc and dac objects.  For the input, 4X oversampling can greatly reduce aliasing noise, and for output we can get noise shaped dither patterns to get effectively about 14 bits a-weighted resolution from the 12 bit DAC.  The oversampled analog input would probably have this:

    "HardwareRequire":[
        {"hardware":"ADC0", "usage":"exclusive"},
        {"hardware":"PDB",  "usage":"176.4kHz"},
    ]

If the normal non-oversampled dac object is already on the canvas, this would be conflict, because the dac uses the PDB hardware as "44.1kHz", but the adcoversample needs it at 176.4kHz.

PaulStoffregen:
I'm planning to add quad channel I2S input/output objects.  A quad channel output would probably have this:

    "HardwareRequire":[
        {"hardware":"I2S_TX0", "usage":"exclusive"},
        {"hardware":"I2S_TX1", "usage":"exclusive"},
        {"hardware":"I2S_MCLK",  "usage":"11.2896 MHz Master"},
    ]

mamuesp:
[...]

PR #130, opened 2015-12-08
Improvement: import is now very error tolerant

mamuesp:
No more need of "begin" or "end" comments - just throw the data in. It searches for keywords (aka node types) at the start of each line and tries to parse these lines.

And yes, the data is completely in the "index.html" file, the resource-folder is obsolete (I forgot to delete it ...) I'm not so happy about this matter because this is a huge amount of data hold as global variables, but the offline paradigm rules. So I can live with that ... ;-)

### DIFF
--- a/gui/index.html
+++ b/gui/index.html
@@ -331,6 +331,52 @@ span.mainfunction {color: #993300; font-weight: bolder}
  TODO: add "parameters" fields, to replace the form html stuff
 -->
 
+<script  type="text/x-red" data-container-name="NodeDefinitions">
+	{"nodes":[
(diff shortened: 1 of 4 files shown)


PaulStoffregen:
did this fully resolve issue #129

mamuesp:
Yes, but without the reference to #59. You may now just throw a full Arduino sketch in the textpanel, and the parser just picks the lines it understands. For the moment I'm looking for a possibility to automatically place the nodes without stacking them (on base of D3 funxtionality like shown [here](http://vowl.visualdataweb.org/webvowl/) or [there](http://bl.ocks.org/MoritzStefaner/1377729)) in one pile when there are no coordinates, after this issue I will look into #59 and affected issues.

mamuesp:
to be a little more discreet, I tried to use the PM channel in the forum ... but this opportunity is blocked ...

Just as information - the tool version online is one with the bug which kills the import function ("count" is not defined) ... perhaps you might install an update.

Regards,
Manfred

PaulStoffregen:
Website updated

PR #136, opened 2015-12-10
Add arm_cortexM4l_math library to linker

ivankravets:
@PaulStoffregen you merge this PR. I see that we have problem with @PlatformIO toolchain where is missed `arm_cortexM4l_math` library. We are working on the new version. I'll reply here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ install:
     # install dependent libraries
     # http://platformio.org/#!/lib/show/539/SerialFlash
     - platformio lib install 539
+    
+    - export PLATFORMIO_BUILD_FLAGS=-larm_cortexM4l_math
 
     # remove "extras" directory
     - rm -r extras


PaulStoffregen:
Still fails, but closer....

    arm-none-eabi-g++ -o .pioenvs/teensy31/firmware.elf -Os -Wl,--gc-sections,--relax -mthumb -mcpu=cortex-m4 -Wl,--defsym=__rtc_localtime=1449756765 -Wl,-T /home/travis/.platformio/packages/ldscripts/mk20dx256.ld .pioenvs/teensy31/src/tmp_ino_to.o -L.pioenvs/teensy31 -Wl,--start-group -lc -lgcc -lm -larm_cortexM4l_math .pioenvs/teensy31/libFrameworkArduino.a .pioenvs/teensy31/libSPI.a .pioenvs/teensy31/libSD.a .pioenvs/teensy31/libSerialFlash_ID539.a .pioenvs/teensy31/libAudio.a .pioenvs/teensy31/libWire.a -Wl,--end-group
    
    /home/travis/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/4.8.3/../../../../arm-none-eabi/bin/ld: cannot find -larm_cortexM4l_math

ivankravets:
@PaulStoffregen we have just used your `arm-none-eabi` toolchain for @PlatformIO and deployed to our storage. Please clear Travis CI cache https://docs.travis-ci.com/user/caching/#Clearing-Caches

Clear Cache

PaulStoffregen:
Getting much closer.  Now only 1 compile fails.  I'm looking at it now to try to figure out why it works on Arduino but fails on platformio.

PaulStoffregen:
Looks like platformio is compiling this code as if it were part of the sketch.  Arduino doesn't.

https://github.com/PaulStoffregen/Audio/tree/master/examples/Synthesis/PlaySynthMusic/miditones

ivankravets:
@PlatformIO builds all source files from the `src` directory by default. There a few options how to change this behaviour: using [src_filter](http://docs.platformio.org/en/latest/projectconf.html#src-filter) or [platformio ci --exclude](http://docs.platformio.org/en/latest/userguide/cmd_ci.html#cmdoption-platformio-ci--exclude).

Nevertheless, I've just removed that folder. See https://github.com/PaulStoffregen/Audio/pull/137

Thanks a lot that help us!

PaulStoffregen:
But there is no "src" directory.

PaulStoffregen:
Arduino's library spec says recursive building is used when there's a "src" folder, and old-style building is used otherwise.

ivankravets:
> But there is no "src" directory.

[platformio ci](http://docs.platformio.org/en/latest/userguide/cmd_ci.html) is shortcut for [platformio run](http://docs.platformio.org/en/latest/userguide/cmd_run.html#cmd-run) command. How does it works? You specify via arguments to [platformio ci](http://docs.platformio.org/en/latest/userguide/cmd_ci.html) which source code you want to build. It can be single file(*.ino, *.c, *.cpp, etc) or path to folder. Then PlatformIO initialise new project from your source files/dirs and place them to [src](http://docs.platformio.org/en/latest/projectconf.html#src-dir) folder. See the original command https://travis-ci.org/PaulStoffregen/Audio/jobs/96062757#L146 and summary
```shell
export PLATFORMIO_CI_SRC=examples/Synthesis/PlaySynthMusic
pip install -U platformio
platformio lib install 539
export PLATFORMIO_BUILD_FLAGS=-larm_cortexM4l_math
rm -r extras
rm -r examples/Synthesis/PlaySynthMusic/miditones
platformio ci --lib=. --board=teensy31

# this is equal to 
platformio ci examples/Synthesis/PlaySynthMusic --lib=. --board=teensy31

# --lib. means here that cloned Git repo need to use as input library for the example
```

If there no SRC arguments to `platformio ci` command, then PlatformIO looks for `PLATFORMIO_CI_SRC` environment. All these feature were specially developed for Continuous Integration systems. They allow to run MATRIX of jobs for the "test" commands.

> Arduino's library spec says recursive building is used when there's a "src" folder, and old-style building is used otherwise.

Ohh... I hope you know my position as for Arduino IDE. Arduino IDE core developers and @adafruit don't like @PlatformIO. I don't have any objections to Arduino IDE. However, I'm not ready to follow these Arduino-styles with code building. I'm trying with maximum to be compatible with Arduino IDE behaviour. Nevertheless, these Arduino "hooks" give bad lessons for C/C++ beginners who start programming with Arduino IDE. In this case, they don't have understanding how code should be organised, what does mean includes and etc.

@PaulStoffregen I hope you understand me. My aim is to create embedded ecosystem where beginner and professional can live comfortable. In a nutshell, PlatformIO looks simple for beginners, like: install platformio, init project, run/upload. From the other side, it can be customised beginning from code builder with own extra scripts and finishing with deployment process via `library.json`, integrating to CI, professional IDE, etc.

---------

Thanks a lot that have spent that time to fix broken builds. As result, we switched to your `arm-none-eabi` toolchain, this is good result from our efforts. I'm not sure that other developers will be helpful for me when I pull similar PR to their repos. I will think on the another solution how to apply continuous integration for the embedded libraries and tests them via multiple boards/platforms. The core idea is to PROPOSE end user who uses @PlatformIO library manager REALLY working library and working examples.

Issue #107, opened 2015-02-19
effect_envelope should do something smarter if noteOff is called before sustain phase


PaulStoffregen:
details here:

https://forum.pjrc.com/threads/27806-Clicking-envelope-with-Audio-Library?p=65273&viewfull=1#post65273

BenjaminPoilve:
I suspect that this is a particular case of this underlying issue but I had an issue (documented [here](https://github.com/BenjaminPoilve/minichord/issues/31)), which happened when:
- An envelope is triggered (noteOn)
- The envelope is called off (noteOff) 
- While in release, a new note is triggered (noteOn), which causes the envelope to decrease per the releaseNoteOn parameter 
- That releaseNoteOn is long enough for the envelope to be called off (noteOff) before the new envelope properly start. 
- It causes the new envelope to "forget" the noteOff and "stick"

I hope that can be useful!

Issue #102, opened 2014-12-13
Reduce cumulative round-off errors in effect_envelope during very slow transitions


PaulStoffregen:
Details here:

http://forum.pjrc.com/threads/27204-Teensy-3-1-Audio-Library-Envelope-limited-to-1000-ms?p=59940&viewfull=1#post59940

PaulStoffregen:
Also this.
https://forum.pjrc.com/threads/28442-Teensy-Drum-Machine-build

PaulStoffregen:
Finally fixed!

PR #142, opened 2015-12-14
Reduce cumulative round-off errors in effect_envelope.

floretan:
Here's a basic implementation of Paul's idea (comment number 12 on https://forum.pjrc.com/threads/27204-Teensy-3-1-Audio-Library-Envelope-limited-to-1000-ms). 

I still need to do some performance comparison, but the code works (tested some 1-2 second attack and release on my synth project, lots of fun). Some improvements can probably be done, any hints are welcome.

This mainly addresses #102, but since longer envelopes states make #107 more likely to be problematic, I also included a very simple fix (just keep the current multiplication factor instead of jumping to the sustain factor).

### DIFF
--- a/effect_envelope.cpp
+++ b/effect_envelope.cpp
@@ -34,6 +34,9 @@
 #define STATE_SUSTAIN	5
 #define STATE_RELEASE	6
 
+// The number of bits added when computing increments.
+#define ENVELOPE_ALIASING 8
+
 void AudioEffectEnvelope::noteOn(void)
 {
 	__disable_irq();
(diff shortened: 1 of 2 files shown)


effect_envelope.cpp
```diff
@@ -125,24 +131,24 @@ void AudioEffectEnvelope::update(void)
 		sample78 = *p++;
 		p -= 4;
 		mult += inc;
-		tmp1 = signed_multiply_32x16b(mult, sample12);
+		tmp1 = signed_multiply_32x16b(mult >> ENVELOPE_ALIASING, sample12);
```
myklemykle: Maybe initialize a variable here, so (mult>>ENVELOPE_ALIASING) doesn't have to happen more than once.  (Unless the compiler already did that.)

PaulStoffregen:
just so I don't forget later, more followup to review here:
https://forum.pjrc.com/threads/27806-Clicking-envelope-with-Audio-Library?p=96459&viewfull=1#post96459

PaulStoffregen:
fixed by https://github.com/PaulStoffregen/Audio/commit/8a5d715dda3903d82d4ebc1aececa40131281014

PR #154, opened 2016-01-17
Just learned that saving r0-r3 is superflues

FrankBoesing:
http://stackoverflow.com/questions/261419/arm-to-c-calling-convention-registers-to-save

### DIFF
--- a/memcpy_audio.S
+++ b/memcpy_audio.S
@@ -44,7 +44,7 @@
 	@ r1: srcL
 	@ r2: srcR
 
-	push	{r3-r12,r14}
+	push	{r4-r12,r14}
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
As I recall, one of the other registers also doesn't need to be saved.  Probably R12.

PaulStoffregen:
Ideally, these should be inline asm rather than function calls.  Then the compiler can save only the registers that really need to be saved, and it chooses the registers best suited for the surrounding code, and the function call overhead is eliminated too.

But inline asm is much more work with gcc's really difficult syntax.

FrankBoesing:
Yes, you're right with inline and i thought about that. But, i'm not familar enough with that syntax :(
About R12: I'm not sure.. but i can try it this afternoon if still works without saving it.

PR #156, opened 2016-01-18
Bug fix so that downward tone sweeps will work.

billpugh:
Fixed problem with tone_hi - tone_lo computation being done
with unsigned arithmetic, and thus always giving a positive result.


PaulStoffregen:
Thanks!

PR #161, opened 2016-01-25
Merge changes from LenShustek/miditones

blackketter:
Includes updates to build on other platforms (including Mac) and non-versioned file name.


PaulStoffregen:
Thanks

PR #170, opened 2016-02-22
faster int32_t substract_32_then_divide (mixed c + assembly)

FrankBoesing:
faster int32_t substract_32_then_divide
absolute worst case: new code 2 cycles faster
best case i've found: new code 30 cycles faster
all my tests showed identical results to the existing function
cycles measured with ARM_DWT_CYCCNT

### DIFF
--- a/synth_dc.h
+++ b/synth_dc.h
@@ -32,20 +32,18 @@
 // compute (a - b) / c
 // handling 32 bit interger overflow at every step
 // without resorting to slow 64 bit math
-#if 0
+#if 1
 // TODO: write this in assembly....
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Looks like this might have a syntax error, line 351 in utility/dspinst.h

PaulStoffregen:
Also, there should be a #ifdef for KINETISK and KINETISL.  Someday I'm going to get the DMA objects working on Teensy-LC.  Many objects will never run on Cortex-M0+, but we can at least fall back to the slower code for these simpler ones.

FrankBoesing:
...works now ;)

PR #190, opened 2016-07-20
important fixes and optimizations

duff2013:
Current state of this library was needing some tlc.

Fixes:
1. NoteFrequency.ino -> wav file sizes where to big, now it will compile.
2. NoteFrequency.ino -> IntervalTimer priority now lower than Audio library.
3. Only 1 Audio block buffer was being used, now both form a ping pong schema.

Speed up processing audio data.


duff2013:
I'm going to close this and redo to pull request, something when wrong with my fork or something I don't reallt know. Not sure what that Travis CI build thing is?

PR #194, opened 2016-08-04
Improve Oversampling for PT8211

Ben-Rheinland:
move OS-related #defines to .h file; speed improvements


PaulStoffregen:
Please change OVERSAMPLING and INTERPOLATION_CIC to something like AUDIO_PT8211_OVERSAMPLING and AUDIO_PT8211_INTERPOLATION_CIC.

When you define them in the .cpp file, they only are visible within that 1 file.  When you put them in the .h file which is included by the main Audio.h header, they become visible to a far wider scope.

Ben-Rheinland:
Done, also changed INTERPOLATION_LINEAR accordingly

PaulStoffregen:
Looks good.  :)

PR #196, opened 2016-08-12
Fix the DC offset computation.

klopyrev:
The previous version would occasionally result in the DC offset getting stuck with the incorrect value. The code relied on the fact that after subtracting the current DC offset s in dc += s / 12000 was offset correctly. However, once an offset become incorrect that assumption wasn't correct and the DC offset would get semi-permanently stuck with the incorrect value (would go on for minutes a time with a large incorrect offset, resulting in RMS and FFT computations being broken).

Here, we just use the average of the last 2048 samples.


klopyrev:
@PaulStoffregen Take a look.

Issue #89, opened 2014-10-12
ADC object should have analogReference()


PaulStoffregen:
http://forum.pjrc.com/threads/26371-Problems-using-Adafruit-electret-mic-amp-and-Audio-library

cosmikwolf:
Hi Paul, just added a pull request for this. I named it setAnalogReference() because there already is a function called analogReference in analog.c.

mcivan:
I ran into this issue today, I like keep up with the current releases, but forgot I have to edit my library. A couple other users have run into this as well. 
https://forum.pjrc.com/threads/39377-A-problem-using-the-Audio-library-and-ADC-library#top

PR #197, opened 2016-08-22
Added ‘setAnalogReference’

cosmikwolf:
to address bug: https://github.com/PaulStoffregen/Audio/issues/89


cosmikwolf:
errmmm perhaps hold off on this for now... It works when you set the ADC pin, but when you use the default constructor it seems to not like the new function. I will test more tomorrow.

PR #201, opened 2016-09-04
Allow play_sd_raw to stream 22050 samples at 44100 for more polyphony

bwhitman:
I love play_sd_raw but want more than a few notes of polyphony in a sample player. So I added a new optional parameter to AudioPlaySdRaw.play(filename, bool halfSR); if it is true the raw files on the SD card are assumed to be at 22050Hz and played back at 44100 by simply doubling each sample. This lets me easily have around 6-7 notes at once from an SD card.

### DIFF
--- a/play_sd_raw.cpp
+++ b/play_sd_raw.cpp
@@ -30,11 +30,17 @@
 
 void AudioPlaySdRaw::begin(void)
 {
+	read_half_SR = false;
 	playing = false;
 	file_offset = 0;
 	file_size = 0;
 }
 
+bool AudioPlaySdRaw::play(const char *filename, bool half_SR)
+{
+	read_half_SR = half_SR;
+	return play(filename);
+}
 
 bool AudioPlaySdRaw::play(const char *filename)
 {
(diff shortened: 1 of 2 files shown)


Pensive:
If you get a better SD card you'll get 6 or 7 at once anyway - but I like the idea, it's a quick and dirty way of doubling polyphony without resorting to a slow-burning ram chip

bwhitman:
@Pensive i have the exact one recommended on this page and only get 2-3 simultaneous 44K RAW files before it glitches, sadly: https://www.pjrc.com/store/teensy3_audio.html

With this patch i can do 6 no problem and maybe 7 before glitching.

FrankBoesing:
I get 6 without problems or patches, with a high-quality sd-card.
But hey, some more would be great. But simply doubling the samples is not good..

bwhitman:
@Pensive @FrankBoesing can you link me your actual microSD card that you are using? I started a thread on the PJRC forums for it: https://forum.pjrc.com/threads/36730-Best-microSD-card-for-reading?p=114323#post114323

PR #205, opened 2016-10-12
Fix distorted USB audio on Macs

dturner:
Recalculated the multiplier and divider values for the I2S MCLK to give exactly 44.1kHz. This means that the audio is no longer distorted on Macs. 

See issue: https://github.com/PaulStoffregen/cores/issues/134


Cinezaster:
Tested it and although it is way beter than before I still has audible distortion (sound stutters) after a couple of minutes (about 6 to 7 minutes). After 12 minutes it corrects itself.
I'm on a Macbook pro 13" running El Capitan
I used a clean synth sound to test and some ambient music. I haven't tested it on my oscilloscope yet.

dturner:
Thanks for testing this. I thought I'd fixed this but on retesting (Macbook Pro 15 running El Capitan 10.11.6 (15G1004)) I also heard glitches, although in my case it was after about 14 minutes, rather than 6 to 7.

PaulStoffregen:
Long term, I believe the only way we're ever going to get mac usb audio working perfectly is switching from async to adaptive mode.  But that imposes a *lot* more work on the Teensy side, not to mention some difficult programming to make all that work happen.  It's on my todo list, but honestly so are many other more urgent things.  I probably won't get to implementing adaptive mode anytime soon.  Sorry.

PaulStoffregen:
One important point about the MCLK settings is the audio lib assumes all the master-mode input and output objects operate in sync.  So the ADC, DAC and PWM objects also need to change to exactly match I2S master, and the USB ones need to be retested carefully on all 3 operating systems (or at least Windows & Linux).

dturner:
That's a shame Paul - I'm still hopeful that async mode could be fixed, since other sound cards (NI Komplete 6) do use this method successfully. As I understand the Mac sends distorted USB audio data when the sync value is above 44.11125 kHz? In theory, with this new MCLK setting there should be no need to have the sync value above 44.110, because the codec should be consuming data at almost exactly 44.100 kHz.

dturner:
Changing the MCLK value doesn't fix the issue, however, having a more accurate MCLK value is a good thing so I've submitted a new pull request here: https://github.com/PaulStoffregen/Audio/pull/213

Issue #209, opened 2016-11-09
The SGTL5000 AutoVolumeControl function does not work


bmillier:
The SGTL5000 AutoVolumeControl function does not work as described. There are 3 incorrect lines in the code which result in the Threshold (16-bit), Attack(12-bit) and Decay(12-bit) registers of the SCTL5000 being loaded incorrectly.

In the file "control_sgtl5000.cpp" the following 3 lines

Line#
844 uint8_t thresh = (pow(10, threshold / 20)*0.636)*pow(2, 15);
845 uint8_t att=(1-pow(10,-(attack/(20*44100))))*pow(2,19);
846 uint8_t dec=(1-pow(10,-(decay/(20*44100))))*pow(2,23);

should be replaced with
uint16_t thresh = (pow(10, threshold / 20)*0.636)*pow(2, 15);
uint16_t att=(1-pow(10,-(attack/(20*44100))))*pow(2,19);
uint16_t dec=(1-pow(10,-(decay/(20*44100))))*pow(2,23);

This loads the registers with the proper 16, 12-bit values, and when I tested it out on the Audioshield, the AVC now works properly.

Pensive:
That explains the funny results I got when I was playing with it. It also explains the crashing I experienced with certain values. Well done sir! Hope to test it soon...

CyborgEars:
I have reported these faulty lines in 2015: https://forum.pjrc.com/threads/29072-Audio-library-control_SGTL5000-cpp-documentation-improvements?p=77939&viewfull=1#post77939

For an in-depth explanation of this issue, read the specification of table 57 to 59 in the SGTL5000 datasheet (rev.6) or my forum entry linked above.

A few days ago, I noticed that, although quite straightforward, the relevant corrections haven't made it into the mainline Audio repository, which is why I've prepared the following commit 9c7671763899a7dcf633f35b847af9b698cc4db8
I did not yet have the time for a thorough review of the possible side effects of this fix on the current code, but as @bmiller has tested this, I have created pull request #210 solving this issue.

@bmillier: did you stumble over my recent commit / the forum post, or did you find this flaw independently?

bmillier:
@CyborgEars    No, I solved it myself- the hard way by examining the code, just like you must have done. I had only been using Teensy for a few weeks at that  time, and hadn't even started to follow the forum or github. I follow the forum now though.

PR #210, opened 2016-11-11
important fix to correct severe AVC threshold/attack/decay issues

CyborgEars:
.. due to truncated variables fed into the SGTL5000 registers.

This fixes issue #209 first discovered in 7/2015, as documented in this [PJRC forum entry](https://forum.pjrc.com/threads/?p=77939#post77939).
Without this fix, the effectiveness of the Automatic Volume Control (AVC) is severely limited, which also compromises it's role as a limiter and safety feature.

### DIFF
--- a/control_sgtl5000.cpp
+++ b/control_sgtl5000.cpp
@@ -840,9 +840,9 @@ unsigned short AudioControlSGTL5000::autoVolumeControl(uint8_t maxGain, uint8_t 
 	if(maxGain>2) maxGain=2;
 	lbiResponse&=3;
 	hardLimit&=1;
-	uint8_t thresh=(pow(10,threshold/20)*0.636)*pow(2,15);
-	uint8_t att=(1-pow(10,-(attack/(20*44100))))*pow(2,19);
-	uint8_t dec=(1-pow(10,-(decay/(20*44100))))*pow(2,23);
+	uint16_t thresh=(pow(10,threshold/20)*0.636)*pow(2,15);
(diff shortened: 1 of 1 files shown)


zirafa:
@CyborgEars, +1 for this patch, tested and works as expected. I could not get AVC to work at all without this fix.

Swap-File:
Is there a reason why this was never merged into master?

CyborgEars:
@Swap-File: I'm also surprised that this is still not merged.

@daspuru 
As far as I can see, the code is still in the same problematic state as five years ago and therefore the AVC behaviour is largely broken, which could be relevant for your (vague) compressor issues:
https://github.com/PaulStoffregen/Audio/blob/5205233a47c7b0a7404157b83e44e9e10922f730/control_sgtl5000.cpp#L845-L847

The original code author @robsoles has commented in the forum post in 2015 that the casting is bad and should be corrected. Multiple people have had to re-discover this problem over the years ( see #209 ) so I don't see why this code should be kept as-is. Since this bug silently breaks the behaviour of the AVC's role as an audio limiter, this is a **safety**-related bug, as outlined before.

@PaulStoffregen , @FrankBoesing : can you take a second look at this issue/PR?

AFAIK, switching to the 16 bit variables has basically no downsides here. Technically, one could add some masking with `0xFFF` for `att` and `dec` to be more compliant with the SGTL5000  the 12 bit register target specification.  (_The upper 4 bits `15:12` are R/O, `0x0` and reserved_).  
But I don't think that this masking is a major concern and the fix already works without it.

PR #215, opened 2016-12-04
Fix BUG for AUDIO_BLOCK_SAMPLES != 128

FrankBoesing:


FrankBoesing:
Don't know why the travis check fails - it compiles with Arduino.

PR #221, opened 2017-01-22
AudioEffectFade should allow instant fades with fadeIn(0) and fadeOut(0)

blackketter:
If you try to fade instantly, it fails as it's trying to divide by zero samples.


h4yn0nnym0u5e:
Very similar fix is in #464, which also fixes other issues

h4yn0nnym0u5e:
> ?? Why not do a PR?

As Paul is in a mood to merge ... I just did

PR #225, opened 2017-04-16
Add waveshaper effect

dxinteractive:
Hi, I wrote a waveshaper effect because I needed it for an expo converter for LFOs, but the effect itself is good for all kinds of things (overdrives, distortions, fuzzes, phase inversion, approximations of many math functions etc.)

Waveshaper input vs output

Here the repo where I was devving it which has some rudimentary docs: https://github.com/dxinteractive/TeensyAudioWaveshaper

First let me know if you're keen to merge it, and if there are any implementation or API changes you'd want me to fix up first.

If that's all good, what else should I provide? Examples? Docs for the gui audio tool? Anything else?

Third, given the time I'd like to make a little web tool to let people draw out their waveshape curves, and return the array of values to load into their sketches. Might also be useful for people wanting to use `WAVEFORM_ARBITRARY` with the `synth_waveform`.

### Example usage

```c++
#include <Audio.h>
#include <Wire.h>
#include <SPI.h>
#include <SD.h>
#include <SerialFlash.h>
#include "effect_waveshaper.h"

AudioSynthWaveform waveform;
AudioEffectWaveshaper waveshaper;
AudioOutputI2S output;

AudioConnection patchCord1(waveform, 0, waveshaper, 0);
AudioConnection patchCord2(waveshaper, 0, output, 0);
AudioControlSGTL5000 audioAdaptor;

int16_t WAVESHAPE_EXAMPLE[17] = {
  -19300,
  -19000,
  -18000,
  -16000,
  -13000,
  -10500,
  -7500,
  -4000,
  0,
  4000,
  7500,
  10500,
  13000,
  16000,
  18000,
  19000,
  19300
};

void setup() {
  AudioMemory(40);
  audioAdaptor.enable();
  waveform.begin(1.0, 200.0, WAVEFORM_SINE);
  waveshaper.shape(WAVESHAPE_EXAMPLE, 17);
}

void loop() {
}
```

### DIFF
--- a/effect_waveshaper.cpp
+++ b/effect_waveshaper.cpp
@@ -1,0 +1,71 @@
+/*
+ * Waveshaper for Teensy 3.X audio
+ *
+ * Copyright (c) 2017 Damien Clarke, http://damienclarke.me
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Most of the audio library API uses floats (0 to 1.0, or > 1.0 for gain) to represent signal levels and attenuation or gain.  I know that would mean requiring 2 copies of the data and allocating RAM... maybe not ideal for this use.  Then again, keeping the API consistent might be worth the cost?

dxinteractive:
I don't mind that idea, it is a little friendlier to look at and I don't think the cost is that great as I assume most usages probably wouldn't require *that* many data points. That also brings it closer in line with the usage of the waveshape effect in the [web audio API](https://developer.mozilla.org/en-US/docs/Web/API/WaveShaperNode/curve), making it one tiny step easier to re-use people's distortion curves (if ever I find any).

On the other hand, `synth_waveform`s `WAVEFORM_ARBITRARY` uses `int_16t`s and it's a little closer to this in terms of how it's used.

Third option is to overload the `waveshaper.shape()` method to accept either. I'm not really a big fan of that as it can make documentation needlessly more difficult to describe.

I'll give it a go using -1.0 to 1.0 values.

dxinteractive:
I've committed that change now, `waveshaper.shape()` now accepts an array of floats instead.

PaulStoffregen:
I started a basic documentation for it.  How's this look?
https://www.pjrc.com/teensy/gui/?info=AudioEffectWaveshaper

dxinteractive:
Hey, yeah looks good for now. It's not the easiest effect to describe so I'm definitely keen to add examples. I've got a couple I'm preparing but not ready just yet. Looks like this PR has been merged, so I've added a couple of missing items to it with this PR: https://github.com/PaulStoffregen/Audio/pull/234. Thanks.

PR #226, opened 2017-04-19
Fix format error: header chunkSize can be 16, 18, or 40

zirafa:
- The CCRMA link no longer works, update link to McGill University
- According to McGill spec, subchunk can be 16, 18, or 40. (Related thread: https://forum.pjrc.com/threads/26889-wav2sketch-problems)


PaulStoffregen:
Thanks!

PR #228, opened 2017-04-22
way faster 32 bit memcpy

FrankBoesing:
32 bit memoryaccess, 32 bit addition, unrolled loop

### DIFF
--- a/output_dac.cpp
+++ b/output_dac.cpp
@@ -137,10 +137,23 @@ void AudioOutputAnalog::isr(void)
 	block = AudioOutputAnalog::block_left_1st;
 	if (block) {
 		src = block->data;
+#if 0		
 		do {
 			// TODO: this should probably dither
 			*dest++ = ((*src++) + 32767) >> 4;
 		} while (dest < end);
+#else		
+		int i = AUDIO_BLOCK_SAMPLES/8;
+		const uint32_t ofs  = (32768 << 16) | 32768;
+		uint32_t * src32 = (uint32_t *) src;
+		uint32_t * dest32 = (uint32_t *) dest;
+		do {					
+			*dest32++ = ((*src32++ + ofs) >> 4) /* & 0x3fff3fff*/; // "&" not needed, DAC is 12 bit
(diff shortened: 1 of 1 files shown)


FrankBoesing:
I'll make an assembler version instead.

PaulStoffregen:
Eventually, I would like to change the PDB to default to 4X the sample rate and bring the oversampling code from PT2811 into the DAC objects.  On the ADC side, simply averaging 4 samples together would be ok.  Eventually, I want to try a small FIR filter on 4X oversampled ADC input.

Any thoughts or ideas on this?  At the very least, it will require a different memory copy, right?

FrankBoesing:
Hi Paul,

please take a look here : 
https://github.com/PaulStoffregen/Audio/blob/master/output_pt8211.cpp#L130
Currently, a memcpy is only used without interpolation.

Some thoughts:

That time Ben wrote this code, I spent some time with trying to optimize 
it - but i could not find a good way. A little bit can be made better 
with unrolling the x4 loop (line 154+) or, maybe, with trying to move 
the conditional out of the loop.

On the other hand, we only need 12 bit, so there might be a way. The 
values used in that code are all 32BIT to minimize errors. With 12BIT 
values this might not be needed and we can use 16 BIT variables instead...

After that, one could try to translate the new code to assembler...



Am 22.04.2017 um 19:33 schrieb Paul Stoffregen:

FrankBoesing:
Hi Paul,

you got me ;-) I have a now working version with 4x oversampling.
There are some questions left..

a) is it ok to create globale define "AUDIO_OUTPUT_OVERSAMPLING" (just 
like AUDIO_BLOCK_SAMPLES)
- this way we can use it for PT8211, too, and it is more clear to the 
user how to use it.

b) isn't the ADC using the same PDB ?

FrankBoesing:
Paul, the compiler makes a pretty good job for the oversampling-code (at least with "mono").
I'll make a fork with all my changes later today - but I'd like that you take a look at the sourcecode before I invest more work on it (-> stereoDACs). 
Then, I'm unsure what to do with the ADCs. Can you make the neccessary changes for the ADCs?

PR #230, opened 2017-04-23
DAC oversampling + memcpy

FrankBoesing:
TODO: AnalogStereo, ADC, PT8211 should use #define AUDIO_OVERSAMPLING


FrankBoesing:
what do you think ?

PaulStoffregen:
Just so you can understand why I haven't acted on this yet... it conflicts with another pending DAC change in a forum thread.  I want to carefully merge both... but haven't had time recently.

PR #234, opened 2017-05-06
Add keywords and include with rest of library

dxinteractive:
A couple of missing items for the waveshaper effect.

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -80,6 +80,7 @@
 #include "effect_delay_ext.h"
 #include "effect_midside.h"
 #include "effect_reverb.h"
+#include "effect_waveshaper.h"
 #include "filter_biquad.h"
 #include "filter_fir.h"
 #include "filter_variable.h"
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
I'm wondering if syntax highlighting "shape" is going to affect people?  Arduino's colors aren't done with any C++ language context.  It just turns the word "shape" orange, even if someone uses this word as a variable.

Maybe there's a longer, more descriptive name that could be used?  Maybe a little more work on examples and a little more text for the design tool documentation might help clarify what this object really does and perhaps a function name to use?

dxinteractive:
Ah, didn't realise that there was no context to the colours, that's a tad annoying. I might change it to `setWaveshape` or something in that case.

I'll commit examples to this PR once their ready, they'll have explanations in comments I can then work upon to describe the effect better in the gui tool. Hopefully won't be too far off.

dxinteractive:
Oh just noticed that this one has been pulled in too, I'll make a new PR with the last set of changes and additions.

PR #244, opened 2017-08-12
Replace DC removal algorithm with IIR HPF, 2 Hz

transmogrifox:
Replaced DC removal code with first order IIR HPF.
Setup as USB sound card and looped analog output to analog input and recorded:
1) Sine chirp to confirm fractional dB roll-off at 20 Hz and flat out to 20 kHz using Audacity.  Spectrogram also looked ok for 12-bit audio.
2) Square wave of 1 Hz to confirm step response settles out to zero.
3) Square wave of 20Hz and 100 Hz to confirm expected exponential decay on each cycle, and clipping looks ok when level set to overdrive.
4) Played music looped-back and inspected waveforms.  There is some wrap-around when clipping but it's hard to know if this is coming from the DAC or the ADC.  Either way the wrap-around doesn't upset the DC removal filter -- doesn't look like things are getting stuck.  Music was up high enough so all was clearly clipping and recovery was good.
5) Played guitar through Teensy with audio pass-through, chorus and flanger.  All sound good.

Guitar has extra audible distortion when original DC removal code is used. Very bad to my ears.  All noise went down to something very "clean" sounding with code update.  I suspect the built-in ADC and DAC+PWM can be applied in a manner capable of competing with the audio add-on, and the DC removal code fixes the non-hardware noise :)

See forum post for graphs of noise in old DC removal code.
Some test programs also can be found here:
http://www.cackleberrypines.net/transmogrifox/src/teensy/dc_removal/

### DIFF
--- a/input_adc.cpp
+++ b/input_adc.cpp
@@ -28,7 +28,7 @@
 #include "utility/pdb.h"
 #include "utility/dspinst.h"
 
-#define COEF_HPF_DCBLOCK    (1048300<<4)  // DC Removal filter coefficient in S7.24
+#define COEF_HPF_DCBLOCK    (1048300<<10)  // DC Removal filter coefficient in S1.30
 
 DMAMEM static uint16_t analog_rx_buffer[AUDIO_BLOCK_SAMPLES];
 audio_block_t * AudioInputAnalog::block_left = NULL;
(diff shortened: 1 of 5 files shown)


transmogrifox:
Deeper analysis revealed 0.005 offset due to exponential error.  Increased resolution to S7.24 to drop this error lower.  The offset is still near -60 dB, so discussion is still open about dropping this below 16-bit resolution.

Because this code uses a 64-bit accumulator I don't see any reason it can't go to S0.31.

transmogrifox:
Went to S1.30 and it appears DC offset is below the 16-bit resolution -- nicely centered on zero and the spectrum looks a little less "hairy" so perhaps some of the low level spectral noise was coming from filter error.

PaulStoffregen:
Thanks for working on this.  The old algorithm was a quick-and-dirty hack that expanded into a bigger hack that never really got carefully tested.

transmogrifox:
Thanks for merging it.  Hopefully I have covered the relevant test cases well enough to avoid creating more issues.  

DC removal is a simple concept yet not entirely trivial.

PR #246, opened 2017-08-26
Add F-RAM CY15B104

FrankBoesing:
Note: 
This chip is a bit demanding - it wants a clean power-on.
I had problems on a simple breadboard, and it was needed to add a capacitor (I use 100uF) to VCC<->GND


FrankBoesing:
thank you !

PR #247, opened 2017-09-01
Adds ByteBeat synthesis object to audio library.

oleiman:
- Adds CPP/H implementation
- updates system design tool HTML
- Includes an example sketch (examples/Synthesis/ProtoByteBeat)

Signed-off-by: Oren Leiman <mumblemumble777@gmail.com>

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -113,5 +113,6 @@
 #include "synth_karplusstrong.h"
 #include "synth_simple_drum.h"
 #include "synth_pwm.h"
+#include "synth_bytebeat.h"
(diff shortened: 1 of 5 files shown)


oleiman:
I cooked this up yesterday and have been finding it fun, interesting, and useful. Glitchy weirdness! This is a basic implementation with several equations adapted from other open source software included by default. I have plans to add facilities for including custom equations at runtime, but I wanted to see whether anybody else gets as much of a kick out of this as I do before going to the trouble.

PR #259, opened 2018-06-12
Introduce pitch functionality in play_serialflash_raw.cpp

visergrass:
This code works fine with a single player in a patch... I'm afraid it will not work properly with more players, because the computation time is about 1500 microseconds, too close to the limit of 2900 microseconds... let's see.
Thanks for the attention


visergrass:
Poliphony doesn't work..  maybe I did some gross mistakes with variables scope?
Thanks

PR #263, opened 2018-09-24
Add control object for TLV32AIC3206

chipaudette:
I've had this for a while from our work on Tympan.  In the Teensy forums, I was asked to share just this module.  So, I unplugged it from the Tympan repo and am offering it here.


control_tlv320aic3206.h
```diff
@@ -0,0 +1,128 @@
...
+#define AIC3206_MIC_BIAS_OFF             0
+#define AIC3206_MIC_BIAS_1_25            1
+#define AIC3206_MIC_BIAS_1_7             2
+#define AIC3206_MIC_BIAS_2_5             3
+#define AIC3206_MIC_BIAS_VSUPPLY         4
+#define AIC3206_DEFAULT_MIC_BIAS AIC3206_MIC_BIAS_2_5
+
+#define BOTH_CHAN 0
+#define LEFT_CHAN 1
+#define RIGHT_CHAN 2
```
PaulStoffregen: Can these 3 macros be moved into the .cpp file (best option), or put inside the class, or at least have "AIC3206_" prepended?

chipaudette:
Great comment regarding the LEFT_CHAN, RIGHT_CHAN, BOTH_CHAN macro names.  I made your suggested revision to add AIC3206 to the macro name.  

I did look at moving these macros into the cpp file, but the macros are there to allow the user to set filters and whatnot to the left and right channels (or both) independently.  So, presumably they need to stay exposed to the user via the h file.  Or, Is there another way that I should have made these settings available to the user (yet still human readable) rather than via #defines?

Otherwise, I renamed the macros to prepend AIC3206 to the macro name.  I also expanded the comments a bit in the h file to show which macros should be used with which of the class's methods.

PaulStoffregen:
Looks fine.  AIC3206_ prefix should be plenty to prevent future conflicts.

PR #268, opened 2018-10-02
memcpy_tdmtx

FrankBoesing:
T3.5, 98 MHz, -O2, Blocksize 128, 1000x calls to memcopy,
in C (original) 10504 microsconds,
asm 6581 microseconds.

Testsketch attached.

[sketch_oct02a.zip](https://github.com/PaulStoffregen/Audio/files/2437658/sketch_oct02a.zip)


FrankBoesing:
added memcpy_tdmrx (input)

PR #275, opened 2018-11-17
Add member restart() to AudioSynthWaveform

ErikDorstel:
The phase(angle) method sets the phase relative to the current angle. A LFO application need to set the phase to an absolute angle. The new method restart() sets the phase_accumulator to zero. The waveform is restarted with an absolute angle defined by phase(angle).

### DIFF
--- a/gui/index.html
+++ b/gui/index.html
@@ -2150,6 +2150,14 @@ The actual packets are taken
 		<a href="http://www.pjrc.com/teensy/td_libs_AudioProcessorUsage.html" target="_blank">AudioNoInterrupts()</a>
 		should be used to guarantee all new settings take effect together.
 	</p>
+	<p class=func><span class=keyword>restart</span>();</p>
+	<p class=desc>
+		Cause the generated waveform to restart its cycle. The generation
+		starts with the angle 0 degrees (default) or with the last value
+		configured by phase(angle);. When multiple objects are configured,
+		<a href="http://www.pjrc.com/teensy/td_libs_AudioProcessorUsage.html" target="_blank">AudioNoInterrupts()</a>
+		should be used to guarantee all new settings take effect together.
+	</p>
 	<p class=func><span class=keyword>pulseWidth</span>(amount);</p>
 	<p class=desc>Change the width (duty cycle) of the pulse.</p>
 	<p class=func><span class=keyword>arbitraryWaveform</span>(array, maxFreq);</p>
--- a/synth_waveform.h
+++ b/synth_waveform.h
@@ -75,6 +75,9 @@ public:
 		}
 		phase_offset = angle * (4294967296.0 / 360.0);
 	}
+	void restart() {
+		phase_accumulator = 0;
+	}
 	void amplitude(float n) {	// 0 to 1.0
 		if (n < 0) {
 			n = 0;


ErikDorstel:
The phase(angle) method sets the phase relative to the current angle. A VCO application need to set the phase to an absolute angle. This is important to get consistent results when multiplying two VCO outputs. The new method restart() sets the phase_accumulator to zero. The waveform is restarted with an absolute angle defined by phase(angle).

skybrian:
Based on the code, I think `phase` currently does something slightly different. It moves the phase by the delta between its argument and the previous call to `phase`. So, if the previous call set it to 2 and this call sets it to 3, then the phase is advanced by 1.

The documentation for phase says something different, though:

> Cause the generated waveform to jump to a specific point within its cycle.

So I think it's a bug, but the description of this patch is also inaccurate about what phase does.

ErikDorstel:
I interpret something else from the code.

ph = phase_accumulator + phase_offset;
// generate audio data
phase_accumulator = ph - phase_offset;
// send audio data

ph represent the angle that is used when generating the audio data.
phase_offset involves shifting the angle. The shift is set regardless of its previous value.
phase_accumulator is the position in the cycle that is passed on from update to update.

Assume the current position in the cycle is 120 degrees. The offset is set to 10, then the angle becomes effective 130 degrees. Assume that the position in the next block is 121 degrees in the cycle and the offset remains unchanged. Then the angle becomes effective 131 degrees. Another block later, at 122 degrees the offset is set to 20 degrees. Then the effective angle becomes 142 degrees.

The phase(angle) member set the phase relative to the current phase.
The restart() member set the phase to 0 plus the offset.

skybrian:
It's not clear to me if we disagree, but I will try to explain better.

By "current position" it seems you mean "the phase stored in phase_accumulator" but this is not the phase being played at the current time. (That is, the time corresponding to the end of the previous block and the start of the next block.)  Whenever we are outside an interrupt, the phase being played is equal to phase_accumulator + phase_offset, what you are calling the "effective angle". This is the phase we will continue with at the start of the next block.

During the interrupt, the ph variable is used temporarily to hold the currently played phase. After the the interrupt is done, phase_offset is subtracted again to uphold the invariant that phase_accumulator + phase_offset is the currently played phase (up to the time just after the current block).

So, phase(angle) sets an offset relative to phase_accumulator, but phase_accumulator doesn't hold the phase currently being played.

If you change the phase_offset from 10 to 20 then the currently played phase moves ahead by 10, but without having the value of phase_accumulator, you don't know what it actually is.

I think the API is confusing because the API user is not controlling the effective, currently played phase, but rather a variable that contributes to it. To change the phase being played we would need to set both phase_offset and phase_accumulator. It's not clear to me that AudioSynthWaveform needs both fields?

ErikDorstel:
That is a good explanation. Mine was a little misleading because of my poor English. I think we have the same understanding of how the code works. I'm not sure why Paul Stoffregen uses two variables for this function. Maybe to simplify the implementation of phase modulation?

My application sets the phase_accumulator to zero with restart() and sets the phase_offset to a certain value with phase(x). So it is possible to control the effective phase.

hb9:
I was looking for a way to bring multiple oscillators back in sync and this MR provided a solution for my problem. After changing my local library accordingly, my problem was solved. I'd love to see this MR merged.

ErikDorstel:
Yes, that is exactly the purpose of this merge request. I have no idea why this has not yet been merged by the maintainer. Maybe I didn't follow the correct procedure?

phjanderson:
> Yes, that is exactly the purpose of this merge request. I have no idea why this has not yet been merged by the maintainer. Maybe I didn't follow the correct procedure?

I am also interested in this functionality. What would it take to get it merged?
I searched but couldn't find anything about procedures for pull requests here.

ErikDorstel:
I submitted this pull request 5 years ago. So far it has not been merged by the maintainer. I don't know the reason. The request does not create any conflicts. Documentation is included.

PR #282, opened 2019-02-05
add input_i2s2, keywords,

FrankBoesing:
+ a little cleanup


FrankBoesing:
nope.. i make a new tomorrow.

PaulStoffregen:
ok, so I can ignore this one?

FrankBoesing:
yes, I'll create a new one this evening, with TDM (at least output for SAI1). More to come this weekend.

jshoek:
I should probably know this by now but when updates are made like this, am I suppose to down load each file listed in the post and place them the library folders or is there an easier way to include all files?

PaulStoffregen:
Yes, to get the very latest you need to pull from github.

Every few weeks I package up an installer with everything that's been merged.  So if you can wait a little while, the installer is the simplest way.

https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=193715&viewfull=1#post193715

jshoek:
I'm assuming with these software updates that they are appropriate for all the 3.x boards.

FrankBoesing:
they are not needed for 3.x boards. they are useful for the teensy-4 only.

PR #283, opened 2019-02-05
Gui T4

FrankBoesing:


FrankBoesing:
Sorry about the reverts.. thought ist better to keep both pullrequests  seperate

PR #284, opened 2019-02-07
various (see below)

FrankBoesing:
input_i2s
input_i2s2
output_i2s2
output_tdm
output_tdm2


FrankBoesing:
current state: See Post 9
https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=193744&viewfull=1#post193744

FrankBoesing:
added MQS

FrankBoesing:
I'd suggest to merge now - It will take some time, several days to add SPDIF, have to read and experiment first, how it works.. 
We can add 3x SPDIF-Out, 2x Software output-only, 1x Hardware with input.

FrankBoesing:
Do we still need i2s_quad?

PR #285, opened 2019-02-11
play .RAW at arb rates

newdigate:
play .RAW from SD card at variable speeds


newdigate:
Apologies, my mistake, I didn't create a feature branch, and by mistake I've included an unwanted commit in this pull request (c796341), which unfortunately I can't seem to rectify, I will create a feature branch and do another pull request...

PR #287, opened 2019-02-16
Play audio sd raw resampled

newdigate:
I've added the ability to play .raw files from sd-card at arbitrary playback rates...  In the future I'd like to extend this to wav files ~~and also allow reverse playback when playback rate is negative..~~.

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -113,6 +113,7 @@
 #include "play_memory.h"
 #include "play_queue.h"
 #include "play_sd_raw.h"
+#include "play_sd_raw_resampled.h"
 #include "play_sd_wav.h"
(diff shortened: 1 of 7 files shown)


PaulStoffregen:
Is the new code MIT license?  I can't consider merging it unless it has the same open source license as we use for the rest of the library.

newdigate:
confirmed, MIT license. I have just added https://github.com/newdigate/Audio/blob/master/LICENSE

newdigate:
Hi Paul. I've added MIT license comments to each source code file I've added. Please let me know if this is acceptable and/or if you have any other suggestions. I have omitted the license from the example file - is that acceptable?

PR #292, opened 2019-03-03
slightly different approach to building with travis

newdigate:
I noticed [PaulStoffregen/Audio](https://travis-ci.org/PaulStoffregen/Audio) is not enabled in travis.  Would you consider turning it on? It looks like it will build now [here](https://travis-ci.org/newdigate/Audio/builds/501074787) 

Another approach might be worth considering which removes dependency on platformio; and discover and compile all .ino files. I think its slightly faster and may be less susceptible to platformio changes. like [here](https://travis-ci.org/newdigate/Audio/builds/501147589)

I'd be really happy to see a travis build badge on your repo. Even a failing build badge is better than no build badge. I'll send a separate pull-request if you're interested.


newdigate:
Hi Frank, I believe it can compile for other teensy boards, thou I haven't tried it. I've also got a few improvements to make - the logging issue with alot of noise from some jaxa.dns. thread is really annoying. Now I've broken the travis build trying to fix the noise on the logging. I'll see if I can get it to build for other platforms, when I get the chance...

newdigate:
Ideally, I'd like to use Arduino boards manager for installing teensy tool-chains, instead of having a separate installer. However, Teensyduino installer needs --dir={arduino.home} argument for headless install- I'd need to send Ardiuno a pull-request allowing arguments to be sent to installers.

newdigate:
Audio library built against teensy4 [travis-t4](https://github.com/newdigate/Audio/tree/travis-t4)

PaulStoffregen:
Previously this CI stuff generated a large number of false alarms.  My time is very limited and divided across so many features people want, and of course just running a small business.  So *any* false alarm is a huge cost.

CI needs to provide enough value to justify its costs.  For complex libraries which target a dozens of chips, catching the cases were change tested on some platforms break builds on others can be pretty compelling.  Testing that the code actually meets some functional test, rather than merely building without compile errors, is also a pretty compelling benefit.

But the cold hard reality is CI that only tests for compile errors has almost no real-world benefit for a library which targets only a few nearly identical hardware configurations.  This library has almost never suffered commits introducing compile errors.  It's just not a likely scenario where the code targets a narrow range of hardware.

While I know some people like to see the build status badge, I do not consider such a cosmetic feature as a CI benefit.

Maybe after we have Teensy 4 released CI might make more sense.  Maybe?  But I'm skeptical.  If this gets merged, and if it generates false alarms and we continue to have a history of commits that compile without error, I will swiftly remove it.  The absolute last thing I need is low-value CI that distracts from important work by generating *any* false alarms.

newdigate:
Yes I agree with most of the points you make; and I understand, and expect your skepticism:
* ci requires continuous effort;  
* compile-only ci has limited benefit;
* unit-testing would be seriously challenging;
* false-positives and false-alarms are annoying and time-waisting. 
* ~~sometimes people love to pimp their github repo with a meanless badge~~

But it's not always about cosmetic badges. It can be symbolic of a mature product, and community; which usually inspires confidence in your brand; And encourages your community by allowing developers to know if their pull-requests will compile, and... if they are breaking more than they are fixing. Also it provides an example for anybody who wishes to integrate their teensy projects on github to build with travis. 

It would be a shame for you or your team to be distracted; But if it costs you nothing, other people are volunteering to do the work to integrate, and you can ignore the failing badge (when it fails), and focus on what you need to, and perhaps be armed with some additional confidence (at least you might know before I send you some broken pull-request), you may find it could save you time. (yep okay not always, but the idea is in the long-term). 

There are ways to minimise the continuous effort of ci, like reducing external dependencies like platform-io, relying only on arduino and teensyduino. (Which, however, takes effort, but like you say, humans will help...)

When I am more confident my build has no false-alarms, or false-positives, (currently latest [here](https://travis-ci.org/newdigate/Audio/builds/501741762) as a matter of interest), I will un-draft this pull request; And you can reconsider, if you are inclined; 

There is no timeline and no pressure, if/when you think this will be beneficial, you can always revisit...

If I break it down to risk / reward / effort equation:

risk=it might distract core developers from spending their time wisely; 
reward=allows developers to be more informed with contributed code
effort=(ideally pjrc team wouldn't waist a moment of time on travis rubbish)

Either way, I am happy, I will try to improve my continuous integration in the mean-time. I thought it was worth a discussion.

PaulStoffregen:
Let's talk of this again in about 8 months.  By then we'll probably be building for more than 1 platform.

PR #294, opened 2019-03-16
AudioPlaySdWav: buffer alignment and irq priority

tsandmann:
Currently there are two problems using AudioPlaySdWav together with the latest version of the SD library:
- for DMA data transfers from the SDHC, the buffer supplied to `File::read()` has to be 4-byte aligned (see https://github.com/PaulStoffregen/SD/pull/17). This PR enforces a 4-byte alignment of `AudioPlaySdWav::buffer` to fix this.
- `AudioPlaySdWav::begin()` disables interrupts completely while calling `SD.open()`. As this includes the *DMA transfer complete interrupt*, `SD.open()` never returns (because `dmaDone` is not updated). As a workaround, this PR only disables interrupts with a lower priority than `SDHC_IRQ`.


manitou48:
the same problem with IRQ and SD.open() occurs in play_sd_raw.cpp

tsandmann:
Pull request updated to cover AudioPlaySdRaw as well.

FrankBoesing:
Any Idea why it disables the interrupts?

manitou48:
I'm not sure why interrupts are disabled.  my hack was to remove the disable_irq() before SD.open(), and that seemed to work for my testing.

https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=207700&viewfull=1#post207700

FrankBoesing:
Hm, Paul?
Is it to prevent other Sd.open(), called by interrupts ?

FrankBoesing:
In this case it is probably better to prevent this in open()

tsandmann:
> Any Idea why it disables the interrupts?

`AudioStream::update()` is called asynchronously from a software interrupt. Its overloads may use `read()` / `write()` on a file handle (as e.g. `AudioPlaySdWav::update()` does) and the filesystem cache of the SD card library is shared between files, but not thread- / interrupt-save IIRC. As (at least) `SD.open()` also uses this cache, software interrupts have to be disabled during `SD.open()`. 

It might be sufficient to just disable `IRQ_SOFTWARE` instead of all interrupts of lower priority than `IRQ_SDHC`, if no other ISR uses SD related stuff, but I didn't double check this.

manitou48:
This bug arose when the SD lib  (SDHC portion) was rewritten to use DMA and DMA interrupt.  But the SD logic in NXP_SDHC.cpp just spins on dmaDone for completion.  dmaDone is set in the DMA ISR.  So another fix would be to change the SD lib to not use DMA ISR, but to spin on the DMA status or SDHC status flag to check for read/write completion. Then no changes would be needed in Audio lib.  The previous version of SD lib polled the SDHC status for completion.

tsandmann:
No longer necessary, it's fixed with 6369a6ae1f3b6ea547a5192f9bf7efe3b778a629.

PR #298, opened 2019-04-23
T4B2: Pin6->Pin7

FrankBoesing:


PaulStoffregen:
Good.  Had this change in my local copy.  Also the RX pin 7->8...

PR #300, opened 2019-05-08
Better initialization of PLL

FrankBoesing:
When experementing with high sample rates for teensyConvolutionSDR, I found that the old mechanic of initialization did not work well in every case. There is a very slight chance that this modified, better code fixes the mclk-resistor-issue too, since it is responisble for the very first MCLKs when the SGTL resets.
The new parameter "force" allows runtime change of the PLL and can be used in sketches.

### DIFF
--- a/utility/imxrt_hw.cpp
+++ b/utility/imxrt_hw.cpp
@@ -24,33 +24,33 @@
  * THE SOFTWARE.
  */
 /*
- (c) Frank b
+ (c) Frank B
 */
 
 #if defined(__IMXRT1052__) || defined(__IMXRT1062__)
 #include "imxrt_hw.h"
 
-//#define CCM_ANALOG_PLL_AUDIO_LOCK	((uint32_t)(1<<31))
-
 PROGMEM
-void set_audioClock(int nfact, int32_t nmult, uint32_t ndiv) // sets PLL4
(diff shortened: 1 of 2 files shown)


PaulStoffregen:
Thanks.  :)

FWIW, just this morning I revised the audio shield PCB to add the 100 ohm resistor.  In a few weeks the next batch goes into production, so in a couple months all the audio boards will have this resistor.

jassing:
Can the 100 ohm be hacked into the current audio shield?

PaulStoffregen:
First, please understand this 100 ohm resistor is only needed with the upcoming Teensy 4.0 board.  It might also help if using Teensy 3.6 and a lengthy wire is used.  But 3.6 works fine without the resistor in most cases.  The older 3.x boards just don't have high enough bandwidth signals for this to matter.

> Can the 100 ohm be hacked into the current audio shield?

The answer really depends on your soldering skills & tools, especially whether you use a microscope.  The MCLK signal is surrounded by GND traces on both sides, so there is no place to easily cut the trace and solder a larger SMT resistor, like a 603 or 805 size.  If you have very good equipment and skills, perhaps you can cut the trace close to the chip and solder a 402 or smaller size resistor across the gap.

Again, this resistor is not necessary for any currently shipping Teensy models.

PR #304, opened 2019-07-10
fix __IMXRT1062__

FrankBoesing:


Defragster:
I tested this to work on the PJRC PT8211 board I built:
https://forum.pjrc.com/threads/54711-Teensy-4-0-First-Beta-Test?p=209881&viewfull=1#post209881

PR #313, opened 2020-02-19
Implement Teensy-LC DAC audio playback

bkurtz:
I mis-read the docs online and was hoping to use a Teensy-LC for audio playback with the prop shield.  Turns out that compiling the Audio library with Teensy-LC is mostly not supported, and that it lacked key functionality for DAC output on Teensy-LC.  This PR fixes the later issue (i.e. finishes implementing DAC output classes), but doesn't fix some of the remaining compile issues (e.g. some Thumb2 assembly instructions in `dspinst.h` and DMA register code designed for KinetisK processors in `output_spdif.cpp`).  In order to compile and use this code, I copied `output_dac.{h,cpp}`, `play_serialflash_raw.{h,cpp}`, and `spi_interrupt.h` from the library directory into my project directory and then modified them so as to avoid pulling in the rest of the audio library.  I would be amenable to making some other tweaks to the audio library so that you can at least *compile* for Teensy-LC with limited functionality - if that's a PR that would be accepted.

Essentially, the work present in this PR consists of implementing the interrupt handler for DMA output to the DAC, changing the setup function that was partially complete to use only a single DMA channel, and tweaking a couple of register settings to the correct values (e.g. needed to activate DMA transfer on timer completion in order to actually activate audio output).  I also changed some things around to use a 2-block output pipeline (`block_left_2nd`) as is done in Teensy 3.x code, even though I didn't see a compelling reason to use an additional block.

### DIFF
--- a/output_dac.cpp
+++ b/output_dac.cpp
@@ -160,14 +160,13 @@ void AudioOutputAnalog::isr(void)
 DMAMEM static uint16_t dac_buffer1[AUDIO_BLOCK_SAMPLES];
 DMAMEM static uint16_t dac_buffer2[AUDIO_BLOCK_SAMPLES];
 audio_block_t * AudioOutputAnalog::block_left_1st = NULL;
+audio_block_t * AudioOutputAnalog::block_left_2nd = NULL;
 bool AudioOutputAnalog::update_responsibility = false;
 DMAChannel AudioOutputAnalog::dma1(false);
-DMAChannel AudioOutputAnalog::dma2(false);
 
 void AudioOutputAnalog::begin(void)
 {
 	dma1.begin(true); // Allocate the DMA channels first
-	dma2.begin(true); // Allocate the DMA channels first
 
 	delay(2500);
 	Serial.println("AudioOutputAnalog begin");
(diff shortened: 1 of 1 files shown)


output_dac.cpp
```diff
@@ -184,59 +183,65 @@ void AudioOutputAnalog::begin(void)
 	// commandeer FTM1 for timing (PWM on pin 3 & 4 will become 22 kHz)
 	FTM1_SC = 0;
 	FTM1_CNT = 0;
-	FTM1_MOD = (uint32_t)((F_PLL/2) / AUDIO_SAMPLE_RATE_EXACT + 0.5);
-	FTM1_SC = FTM_SC_CLKS(1);
+	FTM1_MOD = (uint32_t)((F_PLL/2) / 44117.64706/*AUDIO_SAMPLE_RATE_EXACT*/ + 0.5);
```
bkurtz: This bit would maybe want to change - I was interested in playing back full 44 kHz audio, and in `AudioStream.h` Teensy-LC has been specified to use only 22 kHz.  I didn't see a specific reason not to use the higher rate for my project, so I did that, and I thought it might be good to have a discussion about it here.  But I have no particular attachment to the official library code running at 44 kHz vs 22.  Obviously we don't want the actual sample rate in code here, but the constant isn't part of the audio library, so changing it would be a larger discussion.

FrankBoesing:
? Why not edit AUDIO_SAMPLE_RATE_EXACT?

bkurtz:
> ? Why not edit AUDIO_SAMPLE_RATE_EXACT?

Certainly that seems like the *correct* way to do it.  I guess the main two reasons I didn't do so were
1. I wasn't sure that Teensy-LC contributions for the Audio library were even welcome, since Teensy-LC appears not to be officially supported, and I didn't want to go to extra effort that might just be rejected, and
2. `AUDIO_SAMPLE_RATE_EXACT` is defined *outside* the Audio library, in [AudioStream.h](https://github.com/PaulStoffregen/cores/blob/master/teensy3/AudioStream.h#L62) in the main teensy3 core, which complicates matters a bit.  For example, I'm not sure where all else it might get used, and it seems like changing it there could have significant repercussions for other libraries.

(Also, TBH, I was sortof anticipating some discussion on this point here, and was assuming that we'd probably change it before merging.)

I'll submit another couple of PRs to have more discussion of this and hopefully get things tidied up.

PR #315, opened 2020-03-08
Pwm t4

FrankBoesing:


FrankBoesing:
... increasing the sample-frequency probably should be done for easier filtring.
But I don't have the filters, so that makes no sense for me - i could'nt test it. 
That's a todo for others.

FrankBoesing:
The still may need some work  - don't know.
See my email.

PaulStoffregen:
Ok, I will do some testing here.  Then we can start 1.52 beta and let more people test.

PR #329, opened 2020-03-29
Fix CHIP_I2S_CTRL on Init (SCLK=32*fs -> 64*fs)

FrankBoesing:


FrankBoesing:
For reference: 
Table19.  CHIP_I2S_CTRL 0x0006, Page 32, SGTL5000 Datasheet.

Bit 8: 

> Sets frequency of I2S_SCLK when in master mode (MS=1). **When in slave mode (MS=0), this field must be set appropriately to match SCLK input rate**. 0x0 = 64Fs0x1 = 32Fs - Not supported for RJ mode (I2S_MODE = 1)

PR #330, opened 2020-03-30
async-spdif-input

alex6679:
Implementation of an asynchronous spdif input for the Teensy 4. The input signal is resampled to AUDIO_SAMPLE_RATE_EXACT. A short documentation of the algorithms can be found here: https://github.com/alex6679/teensy-4-spdifIn

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -65,6 +65,7 @@
 #include "analyze_notefreq.h"
 #include "analyze_peak.h"
 #include "analyze_rms.h"
+#include "async_input_spdif3.h"
 #include "control_sgtl5000.h"
 #include "control_wm8731.h"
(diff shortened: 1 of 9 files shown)


FrankBoesing:
Please merge this!

PaulStoffregen:
Ok, merging it now.

But I would like to see an update with defaults for the constructor inputs, so it can be instantiated like all the other objects in the library.

alex6679:
No problem, I'll add default arguments. Another question: There is a very large table that contains the 32bit coefficients of the FIR resampling filter. It is 40961 samples long. Should I declare it as PROGMEM? However, fast access to the coefficients is of course important.

PR #331, opened 2020-04-06
Remove deprecated function and clear code

Paso94:
`arm_cfft_radix4_init_q15` and `arm_cfft_radix4_q15` have been superseded by `arm_cfft_q15` and will be removed in the future from [DSP software library](https://www.keil.com/pack/doc/CMSIS/DSP/html/group__ComplexFFT.html).


MarkTillotson:
You've only changed this in the 1024 fft, not the 256 fft...

Also I'm wondering why not use the real fft, arm_rfft_q15, which is significantly faster than complex fft?

PR #336, opened 2020-04-16
fix compile error in input_i2s2.cpp

manitou48:
fix buffer name to avoid compile error


manitou48:
paul corrected the buffer name for 1.52-beta3

PR #340, opened 2020-05-05
async_spdif_input constructor arguments

alex6679:
default constructor arguments added

### DIFF
--- a/async_input_spdif3.h
+++ b/async_input_spdif3.h
@@ -42,7 +42,7 @@ class AsyncAudioInputSPDIF3 : public AudioStream
 public:
 	///@param attenuation target attenuation [dB] of the anti-aliasing filter. Only used if newFs<fs. The attenuation can't be reached if the needed filter length exceeds 2*MAX_FILTER_SAMPLES+1
 	///@param minHalfFilterLength If newFs >= fs, the filter length of the resampling filter is 2*minHalfFilterLength+1. If fs y newFs the filter is maybe longer to reach the desired attenuation
-	AsyncAudioInputSPDIF3(bool dither, bool noiseshaping,float attenuation, int32_t minHalfFilterLength);
+	AsyncAudioInputSPDIF3(bool dither=true, bool noiseshaping=true,float attenuation=100, int32_t minHalfFilterLength=20);
 	~AsyncAudioInputSPDIF3();
 	virtual void update(void);
 	void begin();


PaulStoffregen:
I'm working on documentation for all the new stuff in the library.

Any chance you could review this page, and suggest descriptions for the 3 functions with "TODO: documentation needed here"?

https://www.pjrc.com/teensy/gui2/?info=AsyncAudioInputSPDIF3

Also, would you like to be mentioned as the author of this input?  If so, please let me know how you'd like your name or alias to appear, and any link to social media or other contact info (if any) you'd like included.

alex6679:
Hi, no problem, I'll post some descriptions here. I'm quite busy right now, but I think I'll have some time over the next weekend. It would be nice if you mention me as the author. You could mention me as Alex Walch. Regarding contact info: I don't use any social media regularly, so it doesn't make much sense to quote a social media contact.

PaulStoffregen:
I've updated the documentation with your name and a link to your github profile.
https://github.com/PaulStoffregen/Audio/commit/a5a42ca0e26e72ae5884244dfe492683c12c1c70

It's also live on the main website design tool page.
https://www.pjrc.com/teensy/gui/?info=AsyncAudioInputSPDIF3

When you have time, please do send descriptions for the 3 functions which are currently documented as "TODO: documentation needed here".

Also the "Notes" section should describe the 4 optional constructor parameters.

alex6679:
Hi,

 

today I found some time to write a few sentences about the asynchronous spdif input. I also had a look at the code and found 1-2 minor bugs and added two functions that return the anti-aliasing attenuation and the resample filter length. I think this functions would be requested once some users start to use the input.

 

Here is the needed documentation:

Feel free to shorten it, if it is too long.

 


getTargetLantency();
Returns the target latency in seconds. The latency is the time from the moment a sample is received by the Teensy SPDIF hardware receiver until it is transmitted by the asrc intput. The audio samples arrive at the asrc input in chunks of 32 samples per channel. The target latency consists of these 32 samples + some buffer that is needed to compensate for timing variations.

 

getBufferedTime();
Returns the buffered time in seconds. The buffered time is the duration of the incoming samples which are not resampled yet. The step width of the resampling algorithm is constantly slightly adjusted to keep the buffered time closely to the target latency. The difference between the target latency and the buffered time is typically smaller than 1 micro second.

 

getInputFrequency();
Returns the sample rate of incoming data, if the PLL has locked, or returns 0 if the audio sample rate is unknown.

 

getAttenuation();
Returns the actual achieved attenuation of the anti-aliasing filter. If the input sampling rate is smaller or equal to 44.1kHz, no low pass filtering is needed and zero is returned.

 

getHalfFilterLength();
Returns the half length of the resampling filter. Its complete length is 2*(the returned value)+1.

 

Notes:
This input is not able to clocks the audio pipline (never has the 'update_responsibility')
The windowed (Kaiser window) sinc-function is used as resample filter (i.e. to interpolate the incoming signal). The longer the filter, the better the quality of the resampled signal. However, a longer filter has a higher group delay and increases the processor usage. The sinc- filter also serves as anti-aliasing filter if the input sample rate is larger than 44.1kHz. The filter length is automatically increased at high input sample rates to reach the specified attenuation. However its half length is restricted to 80. 32bit floating point arithmetic is used at the resampling stage and the resampled signal needs to be transformed to 16 bit integers. Here it is possible to apply triangular shaped dither and noise shaping to increase the perceived signal-to-noise-ratio.

Constructor arguments:
dither: triangular shaped dither is added at the transition from 32bit float to 16bit integer if true (default: true)
noiseshaping: noise shaping is applied at the aforementioned transition if true (default: true)
attenuation: target attenuation of the anti-aliasing filter (default: 100dB). The attenuation is not reached if a filter longer than 161 is needed.
minHalfFilterLength: half of the guaranteed resampling filter (if not longer than 80, default: 20). The filter might be longer if needed to achieve the requested attenuation.


 
 

Gesendet: Sonntag, 24. Mai 2020 um 14:40 Uhr
Von: "Paul Stoffregen" <notifications@github.com>
An: "PaulStoffregen/Audio" <Audio@noreply.github.com>
Cc: "alex6679" <alexander_walch@gmx.at>, "Author" <author@noreply.github.com>
Betreff: Re: [PaulStoffregen/Audio] async_spdif_input constructor arguments (#340)


 

I'm working on documentation for all the new stuff in the library.

Any chance you could review this page, and suggest descriptions for the 3 functions with "TODO: documentation needed here"?

https://www.pjrc.com/teensy/gui2/?info=AsyncAudioInputSPDIF3

Also, would you like to be mentioned as the author of this input? If so, please let me know how you'd like your name or alias to appear, and any link to social media or other contact info (if any) you'd like included.

PR #341, opened 2020-05-13
corrected include path

newdigate:
I wasn't able to build the audio library using cmake because these include paths were incorrect. Once corrected, audio code compiles in both Arduino and using CMake... 

I'd appreciate if you could consider merging this change in to master? Thank you :)


newdigate:
I am no longer getting this issue. Can build remotely using github actions (https://github.com/newdigate/teensy-variable-playback/runs/887024666?check_suite_focus=true)  I'll close the PR...

PR #346, opened 2020-05-28
Gui - requirements

FrankBoesing:
Step #1
Load and check for requirements.
Output - at the moment - on Javascript console.
Step #2: Make conflicts visible (with an Icon), and with a Text-Hint, eventually

Test here: https://forum.pjrc.com/threads/60109-Audio-GUI-for-T4?p=241364&viewfull=1#post241364


FrankBoesing:
Step#2 is a TODO
Maybe I can do in the next days..

PaulStoffregen:
I added a small piece of code to make the error icons appear.
https://github.com/PaulStoffregen/Audio/commit/75121b20648e2073d8e2c1ca92949353c6e3464e

Location of the icon isn't ideal, but still easy to see which object.

The requirementError only seems to be set for newly added objects.  For example, if you start with a blank canvas and add the same input twice, both should get the error icon.  But only the first one get the icon.  I still do not understand why.  Any ideas?

FrankBoesing:
Yes, it needs to be done for both nodes, (much as like setting the errorflags here : https://github.com/PaulStoffregen/Audio/pull/346/commits/dbee3c067760f68397442e24966c0802cea0ebb3#diff-f5e4e46e4439471badb5bad0f28a4cbfR1036)

FrankBoesing:
Best is to add s small function which takes the node ("d" or "r2" as parameter) and adds the icon.

FrankBoesing:
sorry, n2 - not r2

PaulStoffregen:
Please add it and send another pull request.

I need to step away - business stuff - will be back in several hours.

PR #348, opened 2020-05-28
one more fix

FrankBoesing:
ok, the previous issue is now fixed, too.

There is one very minor issue remaining:
If you click "reload"" the first node gets no error-icon.
With the fist click to the workspace it reappears.
I'll do not fix that now.
Maybe one day.

### DIFF
--- a/gui/red/ui/view.js
+++ b/gui/red/ui/view.js
@@ -596,7 +596,7 @@ RED.view = (function() {
 				moving_set.push({n:nn});
 				updateSelection();
 				redraw();
-
+				redraw(); //there is no other way to mark _all_ errors on conflicts :(
 				if (nn._def.autoedit) {
 					RED.editor.edit(nn);
 				}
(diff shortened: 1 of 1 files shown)


FrankBoesing:
I think we're done with that now.

PaulStoffregen:
Wish I could try this right now, but need to wait until tonight.  We're crazy busy here just trying to keep the company running today.

FrankBoesing:
No Problem - keep care.

PaulStoffregen:
Running it now.  Looks really good.  :)

Now it's much easier to tell which objects conflict, because the icon appears immediately on the others when you add the conflicting one.  Very nice.

The next step is the make those icons clickable.  I do not know which GUI way is best to show a message.  Maybe a dialog box?  Or maybe the drop-down panel which is used for other errors?  Or maybe a floating message?  Any ideas?

FrankBoesing:
Not sure what the easiest is.
Didn't look that much at the code.
I from the look&feel I'd prefer a floating message on mouseover.

I can add this, maybe - but not today. I'm busy.. :(


Am 29. Mai 2020 11:01:01 MESZ schrieb Paul Stoffregen <notifications@github.com>:
>Running it now.  Looks really good.  :)
>
>Now it's much easier to tell which objects conflict, because the icon
>appears immediately on the others when you add the conflicting one. 
>Very nice.
>
>The next step is the make those icons clickable.  I do not know which
>GUI way is best to show a message.  Maybe a dialog box?  Or maybe the
>drop-down panel which is used for other errors?  Or maybe a floating
>message?  Any ideas?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/PaulStoffregen/Audio/pull/348#issuecomment-635860378

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #350, opened 2020-05-29
add on-click message

FrankBoesing:


PaulStoffregen:
Using it now.  The notification messages look good.

Also committed a fix for I2S slave output.

PaulStoffregen:
Looks like there's a bug when reloading the page, where 1 of the conflicting objects doesn't get the icon.  But clicking it or the conflicting on causes it to appear.

FrankBoesing:
Yes, I mentioned that in the last PR. At the moment, I don't know how to fix that.

FrankBoesing:
Normally, you don't click "reload".
An then, the missing icon appears if you click anywhere (not only  a node).
So that's a very minor issue.

FrankBoesing:
The updated CSS file is missing on the website, Paul :)

PR #352, opened 2020-05-31
Gui: finally: Fix reload issue

FrankBoesing:
...pls don't forget to copy the css to the website.


PaulStoffregen:
Glad this minor issue is fixed.  :)

I've updated the website including style.css.

PR #353, opened 2020-06-04
Async-spdif-input: SPDIF receiver configuration bug fixed

alex6679:
- Computation of target- and maximum latency fixed
- Order fixed: first the spdif receiver is configured then the dma is started
- Making AsyncAudioInputSPDIF3 a friend class of AudioOutputSPDIF3 -> the same 'config_spdif3()' function can be used across all spdif objects.
- Everything related to the spdif-lock-interrupt is removed/ not needed anymore.
- The interface is extended in order to provide information about the actual filter length of the resampling-filter and the attenuation of the anti-aliasing filter.

### DIFF
--- a/Resampler.cpp
+++ b/Resampler.cpp
@@ -30,7 +30,10 @@
 #include "Resampler.h"
 #include <math.h>
 
-Resampler::Resampler(StepAdaptionParameters settings){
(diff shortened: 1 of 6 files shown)


alex6679:
Unfortunately I had to add another parameter and need to open a new pull-request that includes all changes.

PR #357, opened 2020-07-11
Round down to 0 to avoid the buzzing noise issue

mattvenn:
reported here:
https://forum.pjrc.com/threads/57046-Freeverb-with-low-input-levels-gt-noise-and-wierd-oscillating-quot-tone-quot
Patch by Niels Moseley of http://www.moseleyinstruments.com/

### DIFF
--- a/effect_freeverb.cpp
+++ b/effect_freeverb.cpp
@@ -71,21 +71,27 @@ AudioEffectFreeverb::AudioEffectFreeverb() : AudioStream(1, inputQueueArray)
 	allpass4index = 0;
 }
 
-#if 1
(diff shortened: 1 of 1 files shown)


dcoredump:
Hi @mattvenn 

does your PR mean that the problem also exists in `sat16(n, rshift) signed_saturate_rshift((n), 16, (rshift))` because you are removing this part of the code?

Thanx, Holger

mattvenn:
(From Neils) As far as I can remember, all the saturation routines have the same asymmetrical rounding which causes a buildup of the noise. So, yes, that code was removed on purpose.

Issue #195, opened 2016-08-06
[Feature] Implement MinBLEP Oscillator


DatanoiseTV:
http://www.experimentalscene.com/articles/minbleps.php

> A MinBLEP is a minimum-phase bandlimited step. What is this and why do I want it? Well when you make an oscillator for your synthesizer and you use a mathematical waveform such as square and saw waves you will get aliasing because these waveforms contain harmonics above the nyquist frequency. There are various ways to solve this problem, the main two being oversampling and bandlimited impulse trains (BLITs).

PR #360, opened 2020-07-20
Band limited waveforms

MarkTillotson:
This version of band limited square and sawtooth waveforms is integrated into AudioSynthWaveform and AudioSynthWaveformModulated.

The new tone types WAVEFORM_BANDLIMIT_SQUARE, WAVEFORM_BANDLIMIT_SAWTOOTH and WAVEFORM_BANDLIMIT_SAWTOOTH_REVERSE are used so side-by-side comparisons with originals can be done.


PaulStoffregen:
Is this ready to merge?

zirafa:
Hello, I tested this PR on a Teensy 4.1. I can confirm that the bandlimited waveforms sound much better, thanks for this. The aliasing artifacts are quite audible when modulating non-bandlimited waveforms, but the artifacts disappear when modulating bandlimited waveforms. 

[Here is a test recording I made](https://github.com/zirafa/Audio/raw/bandlimited_waveforms/Test/teensy4.1-bandlimit-test.wav) for comparison:
 
```
- (SINGLE NOTE) reverse sawtooth vs. bandlimited reverse sawtooth
- (SINGLE NOTE) sawtooth vs. reverse sawtooth
- (SINGLE NOTE) squarewave vs. bandlimited squarewave

- (SINGLE NOTE + modulation) reverse sawtooth vs. bandlimited reverse sawtooth
- (SINGLE NOTE + modulation) sawtooth vs. bandlimited sawtooth
- (SINGLE NOTE + modulation) squarewave vs. bandlimited squarewave

- (TWO NOTES + modulation) reverse sawtooth vs. bandlimited reverse sawtooth
- (TWO NOTES + modulation) sawtooth vs. bandlimited sawtooth
- (TWO NOTES + modulation) squarewave vs. bandlimited squarewave
```

---

Note: 
1. The same test using sine and triangle waveforms don't reveal any artifacts when modulating them, so I'm guessing they are already bandlimited in some way. Does that sound right? 
2. My Teensy would occasionally get into an unstable state when switching between the different waveforms during playback. However, the problem is intermittent so maybe it is just a problem with my code. I'll keep poking at it and see if I can reproduce it.

rvh23:
Until I saw this I was about to try and write my own bandlimited oscillator code, which would have been a bit of a task because I'm new to both the Teensy and its Audio library (but not DSP). But I can't find a version of the Audio library that includes these new waveforms. Where should I be looking? Thanks very much, Richard.

zirafa:
@rvh23 You can grab [MarkTillotson's branch referenced in this PR](https://github.com/MarkTillotson/Audio/tree/band_limited_waveforms). Replace the original Teensyduino audio library with this one and then you should be able to use the bandlimited waveforms.

rvh23:
> 
> 
> @rvh23 You can grab [MarkTillotson's branch referenced in this PR](https://github.com/MarkTillotson/Audio/tree/band_limited_waveforms). Replace the original Teensyduino audio library with this one and then you should be able to use the bandlimited waveforms.

Thanks very much. To make matters worse, I'm new to Github too, so I'm yet to figure out how to compile and replace the entire audio library. But in the meantime I just created my own local version of the modified AudioSynthWaveform object (and added the step_table file), and the new bandlimited synthesis seems to work very nicely with a very large improvement over the non-band-limited methods in terms of suppressing aliasing.

rvh23:
>     1. The same test using sine and triangle waveforms don't reveal any artifacts when modulating them, so I'm guessing they are already bandlimited in some way. Does that sound right?

Triangle harmonics fall off much faster than Saw or Square wavs, so the aliasing introduced by simple non-bandlimited synthesis is less obvious, but it's there if you listen carefully. With the Sine in theory there are no higher harmonics, except that the sine in the Audio library is generated using linear interpolation, which produces some distortion that potentially could cause aliasing, but iin practice it's likley to be of little concern.

PaulStoffregen:
Merging this in prep for 1.54-beta6....

krenzlin:
May I ask how the band limited waveforms are generated?
My guess is - just skimming through the code - that it does not use the common PolyBLEP or PTR and instead uses a lookup for the transition and somehow scales it to need.

I am just curious what are the advantages of this method.

Also [this feature request](https://github.com/PaulStoffregen/Audio/issues/195) could be closed with this PR.

PR #374, opened 2021-01-08
Add Teensy LC Code for PT8211

FrankBoesing:


psitech:
Tested and verified to be OK.
See [https://forum.pjrc.com/threads/65735-Teensy-LC-audio-library-support](url)

PR #375, opened 2021-01-13
Fixes for Teensy LC

FrankBoesing:
Quick fixes to avoid compilation errors.

Needed for Blocksize=128.  
Blocksize 128 is needed for the waveplayer.


FrankBoesing:
With this fixes - and blocksize 128 - the waveplayer works on Teensy LC

PR #378, opened 2021-01-21
Fix documentation for SGTL lineInLevel (swap channels)

FrankBoesing:
https://forum.pjrc.com/threads/65905-sgtl5000-ADC-lineInLevel()-Gain-Control?p=267439#post267439


FrankBoesing:
..maybe better than fixing the bug.

PR #380, opened 2021-01-31
Update index.html

manicken:
solves the problem in thread
https://forum.pjrc.com/threads/65954-audio-system-design-tool-bug-with-DAC-and-PWM


PaulStoffregen:
Thanks!

PR #387, opened 2021-02-20
Pdm teensy4  - add support for T4 / IMXRT1062 to PDM input

MarkTillotson:
Needed the relevant setup (similar to input_i2s other than 32 bit word size, and no MCLK/LRCLK pins)
to allow this to function on the T4

### DIFF
--- a/Audio.h
+++ b/Audio.h
@@ -101,6 +101,7 @@
 #include "input_tdm.h"
 #include "input_tdm2.h"
 #include "input_pdm.h"
+#include "input_pdm_i2s2.h"
 #include "input_spdif3.h"
 #include "mixer.h"
 #include "output_dac.h"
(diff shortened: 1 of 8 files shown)


PaulStoffregen:
Agreed, it needs to be 2 different classes to work with the model the design tool uses.

I can merge this now and it'll get changed later.  Or if you want to do it now, just let me know and I'll wait until you say it's ready to merge.

MarkTillotson:
OK I'll think about using a separate class

MarkTillotson:
Ah, see you've merged this - I think its in good shape but haven't been able to test on anything but T4.0 and T3.2
All 6 combinations I've tested:
T3.2  pdm with and without I2S0 output 
T4.0 pdm with and without I2S1 output, pdm2 with and without I2S1 output

PR #398, opened 2021-07-08
add effect_dynamics 

newdigate:
This is a brilliant audio dynamics component from MarkzP [AudioEffectDynamics](https://github.com/MarkzP/AudioEffectDynamics)

One modification from newdigate/AudioEffectDynamics, to calculate gain per sample, instead of per audio block. 

Suggested by @FrankBoesing https://forum.pjrc.com/threads/55117-a-crude-audio-compressor?p=283205&viewfull=1#post283205


newdigate:
I have some corrections to do (Thanks Frank)

make teensy 2.0 friendly using  #ifdef
move log.h to audioeffectdynamics.cpp

PR #399, opened 2021-07-08
Feature/audioeffectdynamics

newdigate:
This is a brilliant audio dynamics component from MarkzP AudioEffectDynamics

One modification from newdigate/AudioEffectDynamics, to calculate gain per sample, instead of per audio block.

Suggested by @FrankBoesing https://forum.pjrc.com/threads/55117-a-crude-audio-compressor?p=283205&viewfull=1#post283205

Update: I've made the code teensyLC friendly, reduced number of files to 2, and change the window size for running rms calc to AUDIO_SAMPLE_RATE / 10 (instead of 4410). Thanks for the suggestions @FrankBoesing


FrankBoesing:
Yes I'd like to see this added :)

Issue #403, opened 2021-08-22
AudioRecordQueue behaves incorrectly with no input


h4yn0nnym0u5e:
### Description

As in title, and discussed at [ https://forum.pjrc.com/threads/68000-AudioRecordQueue-behaves-incorrectly-with-no-input](url)

The Audio library design rules state that a NULL received audio block (from receiveReadOnly()) should be treated as if a silent block had been received. AudioRecordQueue does not do this, but simply fails to count or queue the NULL, even when it has been enabled by queue.begin().

### Steps To Reproduce Problem

Create any audio topology which uses AudioRecordQueue, where the input to the queue can be NULL audio blocks: a good candidate is AudioPlaySdWav. _Without_ calling AudioPlaySdWav::begin(), call AudioRecordQueue::begin() and observe that AudioRecordQueue::available() never reports available (silent) data.

### Hardware & Software

Teensy 4.1 
Audio module
Arduino IDE version 1.8.15
Teensyduino version 1.54
Operating system & version not relevant

### Arduino Sketch

```cpp
#include <Audio.h>
#include <Wire.h>
#include <SPI.h>
#include <SD.h>
#include <SerialFlash.h>

// GUItool: begin automatically generated code
AudioSynthWaveform       waveform1;      //xy=374,336
AudioOutputI2S           i2s;           //xy=566,356
AudioRecordQueue         queue1;         //xy=567,313
AudioConnection          patchCord1(waveform1, 0, i2s, 0);
AudioConnection          patchCord2(waveform1, queue1);
AudioControlSGTL5000     sgtl5000_1;     //xy=558,398
// GUItool: end automatically generated code

void setup() {
  Serial.begin(115200);
  while (!Serial)
    ;
  AudioMemory(10);
}

void loop() {
  int n = 5;
  
  queue1.begin(); // start monitoring
  delay(10);      // should get a few silent blocks
  waveform1.begin(0.5f,261.0f,WAVEFORM_SINE);

  while (n>0)
  {
    if (queue1.available())
    {
      int16_t* qd = queue1.readBuffer();

      for (int i=0;i<128;i++)
        Serial.println(qd[i]);

      queue1.freeBuffer();
      n--;
    }
  }
  while (1)
    ;

}
```

### Errors or Incorrect Output

![image](https://user-images.githubusercontent.com/28339615/130360900-d54fc099-d6cf-49a3-be6b-b0e84b15210a.png)

should look like

[...]

PR #404, opened 2021-08-22
Fix for issue #403 AudioRecordQueue behaves incorrectly with no input

h4yn0nnym0u5e:
record_queue.cpp - don't release() NULL or flag queue entries in ::clear(); allocate and blank a new audio block (if available) in ::readBuffer(), if we have a flag to say NULL (silence)  was received; also keep returning same block until it's discarded by calling ::freeBuffer(); put flag in queue if record is activated by ::begin() but receiving only NULL pointers

record_queue.h - define the "silent data" flag pointer

### DIFF
--- a/record_queue.cpp
+++ b/record_queue.cpp
@@ -43,14 +43,15 @@ void AudioRecordQueue::clear(void)
 {
 	uint32_t t;
 
-	if (userblock) {
+	if (NULL != userblock) {
 		release(userblock);
 		userblock = NULL;
 	}
 	t = tail;
 	while (t != head) {
 		if (++t >= max_buffers) t = 0;
-		release(queue[t]);
+		if (SILENT_BLOCK_FLAG != queue[t]) // real block?
+			release(queue[t]);
 	}
 	tail = t;
 }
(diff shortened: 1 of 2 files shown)


record_queue.cpp
```diff
@@ -59,12 +60,24 @@ int16_t * AudioRecordQueue::readBuffer(void)
...
-	tail = t;
+	if (NULL == userblock) // no block passed to foreground yet
+	{
+		t = tail;
+		if (t == head) return NULL;
+		if (++t >= max_buffers) t = 0;
+		userblock = queue[t];
+		if (SILENT_BLOCK_FLAG == userblock) // got a NULL block, i.e. silence
+		{
+			userblock = allocate(); // allocate a block now, rather than one for every silent entry
```
A-Dunstan, line 71: If allocate() returns NULL the return statement below will crash (NULL dereference).
I think it would be better to have a static empty/silent buffer that gets returned here instead - it may cost a few bytes of permanently allocated memory but would be shared between all recordQueue instances (maybe the audio library already has a const "silence" buffer stashed somewhere for objects to use? If not maybe it should...)

h4yn0nnym0u5e: Good point ... I've pushed a sort of fix. There are issues with statically-allocated audio blocks, because if the reference count goes to zero then the library will attempt to return it to the pool, with less-than-hilarious results.

A-Dunstan: But in this case it's not returning an audio block, it's just returning an int16_t*. So it could definitely return a static buffer.

h4yn0nnym0u5e: True ... but the _right_ fix is to have a static silent audio block known to all audio objects and which behaves properly, rather than an _ad hoc_ permanent waste of 256 bytes (or more, if you've set `AUDIO_BLOCK_SAMPLES` higher) squirrelled away in just one file.

If Paul _ever_ gets around to merging any of these PRs (and after damn near three years I'm beginning to feel it's improbable...) then maybe it'll be worth tweaking the AudioStream core file to do the job properly. For now, with this change, it won't crash!

h4yn0nnym0u5e: I've put this PR in cores: https://github.com/PaulStoffregen/cores/pull/746.

If it ever gets pulled, I'll modify this code (and there are multiple other places in the Audio library which could and should use it, often because they're wasting space having their own private blockful'o'zeroes). 

If not, well, at least now this won't crash, and won't be any worse than it was before, in the marginal case that the programmer hasn't allocated enough blocks for their sketch.

PR #405, opened 2021-08-26
memcpy_audio: 1062 use prefetches

FrankBoesing:


FrankBoesing:
Hm, not worth it.. useful in a corner-case only.

PR #408, opened 2021-09-19
Fix PT8211 case no Audioblocks available

FrankBoesing:
The memset in this case did not help - there was no cache handling because of the following return.  
Also simplified the lines a bit.  
That was a bug, and I hope the fix gets merged,


FrankBoesing:
p.s. it lead to noise i.e. after playing a file.

PR #420, opened 2021-12-07
memcpy_tointerleave allow AUDIO_BLOCK_SAMPLES == 4

ericfont:
Previous code won't work for AUDIO_BLOCK_SAMPLES < 8.  So I would like to add a case for AUDIO_BLOCK_SAMPLES == 4, using the same pattern of interleaving code as for AUDIO_BLOCK_SAMPLES == 8, but just half as much.

I've tested on my Teensy 4.1 with Audio Board Rev D with L/R audio.  I haven't looked into the other two functions memcpy_tointerleaveL and memcpy_tointerleaveR but I assume a similar additional case could be done for that.

### DIFF
--- a/memcpy_audio.S
+++ b/memcpy_audio.S
@@ -85,7 +85,18 @@
 	pkhtb r6,r8,r7,ASR #16
 
 	stmia r0!, {r3,r4,r5,r6}
-	pop	{r4-r8,r14}	
+	pop	{r4-r8,r14}
+#elif AUDIO_BLOCK_SAMPLES == 4
+	push	{r4-r6,r14}
+
+	ldmia r1!, {r5}
+	ldmia r2!, {r6}
+
+	pkhbt r3,r5,r6,LSL #16
+	pkhtb r4,r6,r5,ASR #16
+
+	stmia r0!, {r3,r4}
+	pop	{r4-r6,r14}
 #endif
 	BX lr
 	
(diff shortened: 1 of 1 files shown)


memcpy_audio.S
```diff
@@ -85,7 +85,18 @@
 	pkhtb r6,r8,r7,ASR #16
 
 	stmia r0!, {r3,r4,r5,r6}
-	pop	{r4-r8,r14}	
+	pop	{r4-r8,r14}
```
ericfont, line 88: actually now I'm wondering if that push and pop of r14 is even necessary for (AUDIO_BLOCK_SAMPLES == 8) or for (AUDIO_BLOCK_SAMPLES == 4), since that register is not used in those cases.

ericfont:
hmm...now I'm looking at `memcpy_tointerleaveL` and `memcpy_tointerleaveR`, and I wondering...doesn't `r2` need to be pushed & popped, cause it is overwritten with 0 at https://github.com/PaulStoffregen/Audio/blob/d99090674064a8faa2b1a94f3ff86e27ce8955f2/memcpy_audio.S#L101

memcpy_audio.S
ericfont, line 88: I actually don't believe need to push & pop `r14` for `AUDIO_BLOCK_SAMPLES == 8` or `AUDIO_BLOCK_SAMPLES == 4`.

ericfont:
I can look into modifying memcpy_tointerleaveL and memcpy_tointerleaveR to allow AUDIO_BLOCK_SAMPLES == 4 if that is ok.

ericfont:
closed in favor of #423, which works for L & R only as well as LR, and which also avoids push/pops.

PR #423, opened 2021-12-07
memcpy_tointerleave allow AUDIO_BLOCK_SAMPLES == 4

ericfont:


PaulStoffregen:
looks fine, but now conflicting due to #424

ericfont:
ok, I fixed that conflict and repushed.

PR #432, opened 2022-03-02
T4 pwm output

MarkTillotson:
Fixes some isssues with the T4 AudioOutputPWM - first that cache flushing was done on pointers after they'd been incremented, so I changed the code to index dest[i], rather than use *dest++.  Secondly if another update-responsible class was present update_all could be called twice as there wasn't a guard on the update_all() call in the ISR.

And some extra volatile declarations added where they seemed missing.

I think there's more to do, only the default pins 3&4 are able to be used due to missing begin I think


PaulStoffregen:
Thanks!

PR #433, opened 2022-04-01
Improved AudioEffectDelayExternal

h4yn0nnym0u5e:
- implements a few more options for memory type
- breaks out the memory management into AudioExtMem base class so (future) audio objects can share resources
- documentation added

Subsequent improvements
- resilience against SPI initialisation issues
- speed improvement for SPI-based memory on Teensy 4.x
- update() code simplified by improvements to AudioExtMem class


PaulStoffregen:
oh no, looks like a conflict with another recent commit

h4yn0nnym0u5e:
Think I fixed it - there was a previous PR which corrected a copy-pasta error to do with freeing up delay memory, which hadn't propagated into this one. Or something...

h4yn0nnym0u5e:
Note that https://github.com/PaulStoffregen/Audio/pull/433/commits/054405371eb9f3dc46a5d3302a893ba3d2d7e738 fixes a pretty serious bug that rendered this object almost completely unusable, at least on my Windows 10 / Arduino 1.8.19 system.

h4yn0nnym0u5e:
8ed2b49 reverts by default to the "old scheme" of initialising SPI memory at object construction time, since a fix has been done to the SPI library. Deferred initialisation is still available by setting a third constructor parameter to `false` (if not supplied it defaults to `true`): this may be of use to users who have only partially updated their libraries.

See https://forum.pjrc.com/threads/73154-Teensy-4-1-does-not-boot-when-invoking-SPI-from-external-class and https://github.com/PaulStoffregen/SPI/commit/1904fb5b5449fef66490eba628f6a91dcc80e9da

h4yn0nnym0u5e:
Brought into sync with Teensyduino 1.60 beta 1 updates

h4yn0nnym0u5e:
Added support for 16MB PSRAM fitted to Audio Adaptor.

PR #435, opened 2022-04-13
Fix for problems with SPDIF3

h4yn0nnym0u5e:
- AudioOutputSPDIF3 was flushing cache before writing the memory, so stale data was being DMAed out.
- ADAT had no cache flush (change is untested)
- SPDIF buffer was not being zeroed, so could output brief garbage audio at startup

### DIFF
--- a/analyze_fft1024.cpp
+++ b/analyze_fft1024.cpp
@@ -31,6 +31,7 @@
 
 
 // 140312 - PAH - slightly faster copy
+#if defined(__ARM_ARCH_7EM__)
 static void copy_to_fft_buffer(void *destination, const void *source)
 {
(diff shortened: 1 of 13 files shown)


PaulStoffregen:
Please change this to use arm_dcache_flush() rather than directly manipulating SCB_CACHE_DCCIMVAC.

h4yn0nnym0u5e:
You were too quick doing the merge ... I used arm_dcache_flush_delete(), which was actually a better match for the original code, but appears to be about 1us slower. I've fixed it, but you'd already merged!

PaulStoffregen:
no worries, just send another PR

PR #436, opened 2022-06-04
Compiler warnings in output_pwm.cpp

KurtE:
Paul, not a big deal but when building and debugging I try to remove all of the warnings if possible

There are two warnings in output_pwm.cpp that were caused by the class
defining the member variable block as volatile,

but passed when it was going to release the block, it complained that you are passing it to a non-volatile...

So cast the pointer as to remove the warning.

Sorry sort of combined the warnings fix with maybe fix for PTS8211


h4yn0nnym0u5e:
The PT8211 change has left the call to update_responsibility() in a different position relative to the DMA setup. Should we be concerned?

PR #437, opened 2022-06-06
Actually use arm_dcache_flush() in AudioOutputSPDIF3

h4yn0nnym0u5e:
...as requested, rather than the arm_dcache_flush_delete() which was what the original code effectively had in it. Still seems to work OK, and is faster, too.

### DIFF
--- a/output_spdif3.cpp
+++ b/output_spdif3.cpp
@@ -144,7 +144,7 @@ void AudioOutputSPDIF3::isr(void)
 		// Experience suggests it works better to flush 
 		// after the copy, rather than before...
 		#if IMXRT_CACHE_ENABLED >= 2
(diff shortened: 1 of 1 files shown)


PaulStoffregen:
I saw you mentioned on the forum not having TDM hardware.  I have an extra CS42448 board.  Happy to send it to you.

h4yn0nnym0u5e:
That would be great to have handy for testing. I'm in the UK, if that makes a difference ...

PaulStoffregen:
Not a problem.  PJRC can pay for postal air mail shipping.  But if UK postal service needs to collect a vat-tariff fee upon delivery, you'll need to pay that part.  We can't do it in advance.

Just send me a direct email for details to get this arranged.

h4yn0nnym0u5e:
Done - thanks

PR #439, opened 2022-06-11
SGTL5000 control: support for multiple I²C busses

h4yn0nnym0u5e:
This PR adds a setWire() function to the AudioControlSGTL5000 class, allowing any available I²C bus to be selected to control an SGTL5000 audio shield. Either a reference to an existing TwoWire object or integer index based selection is supported; if the latter is used, then 0-3 map to Wire, Wire1, Wire2 and Wire3.

The GUI design tool documentation and keywords.txt have also been updated.

### DIFF
--- a/control_sgtl5000.cpp
+++ b/control_sgtl5000.cpp
@@ -26,7 +26,6 @@
 
 #include <Arduino.h>
 #include "control_sgtl5000.h"
-#include "Wire.h"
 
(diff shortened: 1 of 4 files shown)


MartijnBraam:
This patch works great for me. It's needed now only the 20-pin version of the SGTL5000 is in stock since that one can't change its i2c address.

PR #446, opened 2022-09-18
Make AudioOutputSPDIF3 sync to AudioInputSPDIF3 to avoid glitches

h4yn0nnym0u5e:
Although AudioInputSPDIF3 is rightly documented as being incompatible with most other audio I/O, it's possible to use its sample clock to drive the AudioOutputSPDIF3 object, making a workable S/PDIF pass-through possible without the need to resample, 

In addition to the required code changes, the design GUI has been updated so it doesn't show a conflict between the two SPDIF3 objects, and the info pane text has been edited to give a little more information on usable combinations of I/O objects with these two.


h4yn0nnym0u5e:
Doesn't do what I thought! Looking into it...

h4yn0nnym0u5e:
OK, so this now achieves its original aim of syncing AudioOutputSPDIF3 to the AudioInputSPDIF3 sample rate, avoiding the overhead of resampling. It also adds clocking of I2S1 / SAI1 from the S/PDIF input sample rate, tested at 32, 44.1, 48 and 96kHz - the update rate changes to match, so anything frequency-related will be more-or-less wrong, depending on how different the S/PDIF rate is from AUDIO_SAMPLE_RATE_EXACT.

_However_, for some reason there's an extra or dropped sample on I2S every so often, despite the shared clock source. I can't find this to fix it... I've toggled an output pin in both the I2S and SPDIF ISRs, and they're definitely locked together.

On this basis, I've updated the Design Tool to permit use of I2S with AudioInputSPDIF3 without flagging incompatibility, and put some comments in the info pane relating to "minor degradation" of the I2S audio quality.

h4yn0nnym0u5e:
Closing this in favour of a new, more capable version.

PR #448, opened 2022-10-29
Make AudioPlaySdWav object honour the value of AUDIO_BLOCK_SAMPLES

h4yn0nnym0u5e:
Loads either one SD card sector / 512 bytes, or two blocks' worth (to allow for stereo playback) if AUDIO_BLOCK_SAMPLES is set to >128. Minor change to play_sd_raw.cpp to remove magic number.

Very simple update, no (intentional!) change in behaviour with standard library settings.

### DIFF
--- a/play_sd_raw.cpp
+++ b/play_sd_raw.cpp
@@ -96,7 +96,7 @@ void AudioPlaySdRaw::update(void)
 
 	if (rawfile.available()) {
 		// we can read more data from the file...
-		n = rawfile.read(block->data, AUDIO_BLOCK_SAMPLES*2);
+		n = rawfile.read(block->data, AUDIO_BLOCK_SAMPLES*sizeof(int16_t));
 		file_offset += n;
 		for (i=n/2; i < AUDIO_BLOCK_SAMPLES; i++) {
 			block->data[i] = 0;
--- a/play_sd_wav.cpp
+++ b/play_sd_wav.cpp
@@ -167,11 +167,11 @@ void AudioPlaySdWav::update(void)
 		if (consume(n)) return; // it was enough to transmit audio
 	}
 
-	// we only get to this point when buffer[512] is empty
+	// we only get to this point when buffer[] is empty
 	if (state != STATE_STOP && wavfile.available()) {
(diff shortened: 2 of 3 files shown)


dwexel:
This is pretty much what I came to github to find, except I wanted to remove the magic number in play_sd_wav.h
It should reference AUDIO_BLOCK_SAMPLES instead of being a hard-coded 512 bytes
I don't know anything about SD card sectors.

h4yn0nnym0u5e:
> This is pretty much what I came to github to find, except I wanted to remove the magic number in play_sd_wav.h It should reference AUDIO_BLOCK_SAMPLES instead of being a hard-coded 512 bytes I don't know anything about SD card sectors.

But it's necessary to understand SD card sectors _and_ the `AUDIO_BLOCK_SAMPLES` value _and_ the contents of .WAV files to make a meaningful change here. I agree that the use of a magic number is inherently poor, but simply reading `AUDIO_BLOCK_SAMPLES * 2` to get enough for one stereo `update()` (or two mono ones) is very inefficient for small block sizes (often used by those who want less than the 2.9ms latency given by the stock values). Hence the slightly more complex calculation in this PR.

If you've successfully used this in your code, please report back - I'd like to know it works for you, even if at nearly 3 years old the prospect of it getting pulled in seems remote. Do note that you're unlikely  get more than a couple of files playing back at once, as the design of this object is pretty poor, accessing the SD card inside an interrupt.

dwexel:
If you were to redesign this object from the ground up, what would that look like?
If that's something you want to do, I'm interested in trying to help.

h4yn0nnym0u5e:
Very much like [this PR](https://github.com/PaulStoffregen/Audio/pull/483). Yes, it uses more memory, and needs a bit of extra setup, but it seems to work well. Precautions are needed around other SD access, because SdFat calls `yield()` internally, which can result in crashes if there's a pending EventResponder response which itself reads or writes the SD. 

See also the [discussion thread here](https://forum.pjrc.com/index.php?threads/yet-another-file-player-and-recorder.70963/).

PR #449, opened 2022-11-14
Fix issues with AudioSynthWavetable object

h4yn0nnym0u5e:
- various places didn't check for NULL pointers, and cause a crash if used incorrectly*
- "documentation" consisted largely of "blah blah"

\* e.g. calling setFrequency() on an object that's not actually playing. Apparently you could get away with this on a teensy 3.2, but no longer, if you use a Teensy 4.x. See https://forum.pjrc.com/threads/71479-Teensy-4-0-audio-shield-spontaneously-resets


PaulStoffregen:
Thanks!

PR #450, opened 2022-12-18
A simple stereo out sample player 

timnbraun:
With optional phase inversion per channel.


PaulStoffregen:
Why does this create another play memory class?

PR #451, opened 2022-12-29
Play_raw Play_wav to any FS

KurtE:
@PaulStoffregen @mjs513 @h4yn0nnym0u5e

This is an alternate implementation of
#419

As mentioned in the forum thread: https://forum.pjrc.com/threads/71716-recording-playing-back-audio-on-multiple-SD-cards

It simply adds optional parameter at the end of the play method for the FS which defaults to SD

The other PR, makes a copy of the play method and adds the FS as the first parameter.

Either way works for me.

### DIFF
--- a/play_sd_raw.cpp
+++ b/play_sd_raw.cpp
@@ -37,8 +37,9 @@ void AudioPlaySdRaw::begin(void)
 }
 
 
-bool AudioPlaySdRaw::play(const char *filename)
+bool AudioPlaySdRaw::play(const char *filename, FS *pfs)
 {
+	if (!pfs) return false;
 	stop();
 #if defined(HAS_KINETIS_SDHC)
 	if (!(SIM_SCGC3 & SIM_SCGC3_SDHC)) AudioStartUsingSPI();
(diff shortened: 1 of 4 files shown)


h4yn0nnym0u5e:
Any reason this is using FS* as the parameter rather than FS& ? The latter would seem to be safer, as there's then no need for a NULL pointer check (which you don't make...)

KurtE:
Could go either way, my first version did use references.  My first usage was with array with pointers, so had to dereference them passed in.  Actually both classes do test for null pointer Sent from my iPadOn Dec 31, 2022, at 2:41 PM, Jonathan Oakley ***@***.***> wrote:﻿
Any reason this is using FS* as the parameter rather than FS& ? The latter would seem to be safer, as there's then no need for a NULL pointer check (which you don't make...)

h4yn0nnym0u5e:
> Could go either way, my first version did use references.  My first usage was with array with pointers, so had to dereference them passed in.  Actually both classes do test for null pointer Sent from my iPad

_Mea culpa_, I didn't inspect your code properly - indeed you do have a check for NULL pointers. As an "old school C programmer" I'm not sure if there's a "preferred C++ way" ... but moving it from a run-time to a compile-time check feels better to me, as you won't get unexpected silence from a NULL pointer, or weird crashes if you somehow manage to pass an invalid pointer (a skill I seem to have mastered...).

KurtE:
Closing out detritus

PR #470, opened 2024-01-18
Added multi-pin TDM support.

ccrome:
This allows you use use more than one TX or RX pin in TDM mode, allowing for up to 64 channels of record or playback data

This fix goes hand-in-hand with the multi-channel USB support. https://github.com/PaulStoffregen/cores/pull/732

input_tdm: every odd channel had every other sample swapped In every odd channel in TDM input (1, 3, 5, 7, 9, 11, 13, 15), every other word was swapped due to an incorrect copy from 32-bits to
16-bits. This fix corrects the odd channels.   The Shift-by zeros  and the
extraneous logical ands are there for clarity, and I verified they
don't end up affecting final code optimization as long as optimization
is turned on.


h4yn0nnym0u5e:
This is non-functional, as the changes to TDM output are not complete

PR #472, opened 2024-02-17
Fix output_tdm2 excessive copy

h4yn0nnym0u5e:
`memcpy_tdm_tx()` waas copying twice as much data as it should, overflowing the output buffer. Now does the same as `output_tdm`, which appears to be correct.


h4yn0nnym0u5e:
Confirmed in forum post: https://forum.pjrc.com/index.php?threads/bug-in-audiooutputtdm2.74502/post-344714

jabeka:
Confirmed here too ! FIxed non working outptut on TDM2 while the same hardware works like a charm on TDM. Now works flawlessly on both.

PR #475, opened 2024-04-30
added method for restarting the phase accumulator

phjanderson:
Added a method restart() to AudioSynthWaveform and AudioSynthWaveformModulated to reset the phase_accumulator to 0. This allows the phase of multiple oscillators in a synthesizer to be synchronized upon note on.

### DIFF
--- a/gui/index.html
+++ b/gui/index.html
@@ -3341,6 +3341,14 @@ The actual packets are taken
 		<a href="http://www.pjrc.com/teensy/td_libs_AudioProcessorUsage.html" target="_blank">AudioNoInterrupts()</a>
 		should be used to guarantee all new settings take effect together.
 	</p>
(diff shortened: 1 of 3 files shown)


chipaudette:
I am completely butting my nose in here...but I support the idea that this is a worthy functionality to include.  Oscillator reset is absolutely a thing.  It's not super-common, but those synths that do it are very distinctive and iconic.  Moog Taurus bass pedals being probably the most iconic.

So, yeah, I have not tried this particular implementation, but the motivation is good.  It's a worthy functionality for an oscillator to have.

Chip

ErikDorstel:
This is an absolutely necessary function, for example to start modulating the frequency or amplitude with an LFO with a fixed phase angle at every "note on".

I've been using it for years.

https://github.com/PaulStoffregen/Audio/pull/275

PR #476, opened 2024-06-01
fix MidiSynthLarge example to work on Teensy4

A-Dunstan:
MidiSynthLarge example uses too much wavetable data for Teensy 4's DTCM - put large const data arrays in PROGMEM to fix it


PaulStoffregen:
Thanks!

Issue #479, opened 2024-09-04
Error in input_tdm.cpp and input_tdm2.cpp


palmerr23:
### Description

When a device that uses 16 bit TDM slots is connected, input data is corrupted on the all odd numbered channels. The issue is not revealed with 32-bit slots as the odd channels are ignored.

This is because of an incorrect calculation on line 102:

` *dest2++ = (in1 << 16) | (in2 & 0x0000FFFF);`

This code swaps the samples, putting the first sample in the high part, and the second sample in the low part of the 32-bit destination word. This is opposite to the behaviour of dest1 on line 101:		

  `*dest1++ = (in1 >> 16) | (in2 & 0xFFFF0000);`

**the solution is to substitute line 102 with this line of code in both files:**

  `*dest2++ = (in1 & 0x0000FFFF)  | (in2 << 16);`

The solution has been tested on CS42448 and TLV320AIC3104 hardware in TDM mode.

### Steps To Reproduce Problem

16-bit TDM hardware is required, such as my new 4xTLV320AIC3104 board. 

Any code that inputs L&R TDM channels will work.

### Hardware & Software

Board 
  T4.0
Shields / modules used New 
  TLV320AIC3104 board
Arduino IDE version 
  2.3.2
Teensyduino version 
  1.59
Operating system & version 
  Win11
Any other software or hardware?

### Arduino Sketch

#include <Audio.h>
#include <Wire.h>
#include <SPI.h>
#include <SD.h>
#include <SerialFlash.h>
#include "control_tlv320aic3104.h"

#define BLOX 40
AudioInputTDM             tdm1;           
AudioOutputUSB           usb1;  
AudioConnection          patchCord15(tdm1, 0, usb1, 0);
AudioConnection          patchCord16(tdm1, 1, usb1, 1);
AudioControlTLV320AIC3104 aic;

void setup() 
{
  AudioMemory(BLOX);
  Wire.begin();

  aic.audioMode(4, true, AICMODE_TDM);
  aic.inputMode(AIC_SINGLE);
  aic.enable( ); 
  aic.gain(0); 
}

void loop() 
{
}

### Errors or Incorrect Output

Right channel input shows a spike at (22.05 -f) kHz, with an amplitude that varies with the input.

Left
![left_channel](https://github.com/user-attachments/assets/06010c24-4843-4e97-a67c-d7dc18c0a072)

Right
[...]

PR #480, opened 2024-09-04
Fix odd channel sample swap for TDM inputs

palmerr23:
In the existing code, the samples for odd channels are swapped when being transferred from the DMA buffer to the audio buffer.

dest1 is correctly calculated, however the early sample for dest2 goes into the high part of 32-bit *dest2, and the later sample goes into the low part.

This is not obvious for 32-bit TDM slots as the advice is to ignore odd TDM inputs. For 16 bit slots it results in corruption - see Issue 479 [https://github.com/PaulStoffregen/Audio/issues/479](url)
```
*dest1++ = (in1 >> 16) | (in2 & 0xFFFF0000);
*dest2++ = (in1 << 16) | (in2 & 0x0000FFFF);
```
the correct code is 
```
*dest1++ = (in1 >> 16) | (in2 & 0xFFFF0000);
*dest2++ = (in1 & 0x0000FFFF)  | (in2 << 16);	
```	
The changed code has been tested on both CS42448 and TLV320AIC3104 hardware in TDM mode.

### DIFF
--- a/input_tdm.cpp
+++ b/input_tdm.cpp
@@ -99,7 +99,7 @@ static void memcpy_tdm_rx(uint32_t *dest1, uint32_t *dest2, const uint32_t *src)
 		in2 = *(src+8);
 		src += 16;
 		*dest1++ = (in1 >> 16) | (in2 & 0xFFFF0000);
-		*dest2++ = (in1 << 16) | (in2 & 0x0000FFFF);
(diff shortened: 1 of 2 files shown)


palmerr23:
See PRs 440 and 480

palmerr23:
PR 440 addresses the same issue and has been merged.

PR #481, opened 2024-09-11
Some minor fixes

h4yn0nnym0u5e:
- effect_envelope.cpp: don't re-start release if noteOff() called in release phase
- effect_granular.h: add guard macros
- effect_reverb.cpp: don't just return on NULL block received, process silent data (otherwise the reverb cuts off in a nasty fashion...)


ssilverman:
May I suggest a slightly better commit message? For example, “Some minor fixes for effects”.

PaulStoffregen:
@h4yn0nnym0u5e - Going to look at this and many other audio things in the coming weeks.  But just to let you know where things are right now, my build process for the Windows installer broke in a couple ways.  Signing is the big problem, which used to involve only files but now requires a hardware token which isn't working with the virtual machine I previously used.  I got it up and running again for 1.60-beta1, but on a very slow Intel NUC mini PC.  I have a faster machine on order, should arrive in about 1 week.  The UPX compression is also no longer working for Windows (seems we've hit its internal size limit), and for all the other platforms compressing is terribly slow, so today I'm working on replacing easy-but-limited UPX with using ZSTD internally.  Retooling my build process isn't the fun way I wanted to spend this weekend, but I've been putting it off for years and now things are reached the breaking point (rather than just slowing me down).  I really need to get these tooling problems solved before I work on merging a lot of stuff.

h4yn0nnym0u5e:
Sure, understood. I do appreciate you have other calls on your time, plus of course when you can't even do your _planned_ things because the tools have mysteriously decided to break themselves. Been there, done that...

I'm trying to spend a bit of time getting anything I think merits consideration for TD 1.60 "up to speed" so at least there aren't any surprise merge conflicts, and all the examples and documentation are done. Obviously it'd be really handy to have some indication of what you'd like to include, if you have a couple of moments to share that.

PR #482, opened 2024-09-13
Support for PCM3168

sonalux:
Adding support for the PCM3168 codec that should replace the deprecated CS42448


h4yn0nnym0u5e:
Why are you submitting a PR for code that I wrote?

PR #485, opened 2024-11-18
(draft) Add implementation of AudioWindowKaiserBessel

scrain777:
This PR implements the KaiserBessel window from https://en.wikipedia.org/wiki/Window_function and adds the 
computed window to the available windows.  The Python script used to generate the windows is included alongside the Perl script used to generate other windows.

KaiserBessel has a shape parameter. I selected one specific value (3) because it was excellent for my application. In particular, I was wanting to minimize the width of the leakage at the -60dB level. The Python script can be trivially modified to implement additional variants of KaiserBessel if that is desired.


gui/index.html
```diff
@@ -4999,7 +5000,7 @@ <h3>Notes</h3>
 		<li><span class=literal>AudioWindowWelch1024</span></li>
 		<li><span class=literal>AudioWindowHamming1024</span></li>
 		<li><span class=literal>AudioWindowCosine1024</span></li>
-		<li><span class=literal>AudioWindowTukey1024</span></li>
```
scrain777, line 5002: This line is not supposed to be deleted.

scrain777:
I have not tested this code yet.

scrain777:
> I have not tested this code yet.

Tested with the FFT example on a Teensy 3.5. Works great, although on the sine example the Hanning window has less leakage....

scrain777:
Sorry, I closed the PR by mistake.

scrain777:
Also tested in a real application on a Teensy 4.1 with audio shield.

